PR #23, opened 2020-12-09
Implement client base and storage endpoints

fnesveda:
I've implemented the base of the client and the storage endpoints. It's modeled after the javascript client so the usage is similar. There's still a lot of documentation missing, and I need to implement some integration tests for these, but it should be good as a start of the project.


src/apify_client/_utils.py
```diff
@@ -0,0 +1,101 @@
...
+        nonlocal swallow
+        swallow = False
+        raise exception
+
+    for attempt in range(1, max_retries):
+        try:
+            return func(bail, attempt)
+        except Exception as e:
+            if not swallow:
+                raise e
```
mnmkng, line 83: I like that you implement the exp backoff using a loop. I wanna do that too later, to clean up the stack trace and get rid of a dependency.

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -0,0 +1,227 @@
...
+
+        return {
+            'items': data,
+            'total': int(response.headers['x-apify-pagination-total']),
+            'offset': int(response.headers['x-apify-pagination-offset']),
+            'count': len(data),  # because x-apify-pagination-count returns invalid values when hidden/empty items are skipped
+            'limit': int(response.headers['x-apify-pagination-limit']),  # API returns 999999999999 when no limit is used
+        }
+
+    def download_items(
```
mnmkng, line 112: I don't know what the convention is in Python, but I kinda dislike functions that return different types based on some configuration parameter. With `getRecord` I used `stream` and `buffer` as options, because the function doesn't really have any other options. But in case of `downloadItems`, I was thinking of actually creating a different function for streaming, like `createItemStream` In Node, you could then 

```js
createItemStream('csv').pipe(fs.createWriteStream('data.csv'))
```

I think we should agree on this, so that the API stays the same across the clients.

mnmkng (commented):
I obviously did not read the whole thing, because, well, I can't read Python very well. Great job @fnesveda anyway! This is a start of something great for Apify (and the world).

src/apify_client/_utils.py
```diff
@@ -0,0 +1,101 @@
...
+        return cast(Dict, parsed_response['data'])
+
+    raise ValueError('The "data" property is missing in the response.')
+
+
+T = TypeVar('T')
+BailType = Callable[[Exception], None]
+
+
+def _retry_with_exp_backoff(
```
mhamas, line 60: What about turning this in decorator? I think it would make it less verbose and easier to use.

Decorator example (for iterating function N times, I chose this as it's not that long and illustrates the point).
```
from typing import Any, Callable
from functools import wraps


def iterate_decorator(num_of_iterations: int) -> Callable:
    def decorator(func: Callable) -> Callable:
        @wraps(func)
        def iterate_fn(*args, **kwargs) -> Any: 
            for attempt in range(num_of_iterations):
                print(f"Iteration {attempt} in decorator")
                func(*args, **kwargs)
        return iterate_fn
    return decorator


@iterate_decorator(num_of_iterations = 3)
def print_foo_decorated(arg: str, kw_arg: str):
    print('foo decorated')


print_foo_decorated('arg', kw_arg='kw_arg')
```

When executed, this returns
```
Iteration 0 in decorator
foo decorated
Iteration 1 in decorator
foo decorated
Iteration 2 in decorator
foo decorated
```

src/apify_client/_http_client.py
```diff
@@ -0,0 +1,108 @@
...
+        params: Dict = None,
+        data: str = None,
+        json: Union[str, Dict, List[str], List[Dict]] = None,
+        stream: bool = None,
+        parse_response: bool = True,
+    ) -> requests.models.Response:
+        request_params = self._parse_params(params)
+        requests_session = self.requests_session
+
+        def _make_request(bail: Callable, attempt: int) -> requests.models.Response:  # type: ignore[return]
```
mhamas, line 35: We are currently forcing any retryable function to have `bail` and `attempt` args an no other args. I don't think this is very flexible.

src/apify_client/_utils.py
mhamas: Also, I'd not force decorated function  to have `bail` and `attempt` arguments. I think they should be able to have any arguments. Instead, I think we should pass to decorate itself on which exceptions it should continue and on which it should simply throw. Let's abstract that from the function decorated.

src/apify_client/client.py
```diff
@@ -0,0 +1,90 @@
...
+            min_delay_between_retries_millis: How long will the client wait between retrying requests (increases exponentially from this value)
+        """
+        self.token = token
+        self.base_url = base_url
+        self.max_retries = max_retries
+        self.min_delay_between_retries_millis = min_delay_between_retries_millis
+
+        self.http_client = _HTTPClient()
+        # TODO statistics
+        # TODO logger
```
mhamas, line 40: nit for the future. Let's use `logging` package for logs.

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -0,0 +1,10 @@
+from .resource_client import ResourceClient
+
+# from ..._utils import _parse_date_fields, _pluck_data
```
mhamas, line 3: forgotten from experiments?

src/apify_client/clients/base/base_client.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any, Dict
+
+from ..._http_client import _HTTPClient
+from ..._utils import _to_safe_id
+
+
+class BaseClient:
+    """Base class for sub-clients."""
+    def __init__(self, *, base_url: str, http_client: _HTTPClient, resource_id: str = None, resource_path: str, params: Dict = None) -> None:
```
mhamas, line 9: Nit: `resource_id` and `resource_path` types should be `Optional[...]`

src/apify_client/clients/base/base_client.py
```diff
@@ -0,0 +1,37 @@
...
+            resource_id (str): ID of the manipulated resource, in case of a single-resource client
+            resource_path (str): Path to the resource's endpoint on the API server
+            params (dict): Parameters to include in all requests from this client
+        """
+        self.base_url = base_url
+        self.http_client = http_client
+        self.params = params or {}
+        self.resource_path = resource_path
+        self.resource_id = resource_id
+        self.url = f'{self.base_url}/{self.resource_path}'
```
mhamas, line 24: Can we enforce that `url` never ends with `/`? Otherwise it would start making problems due to double `/` (saw this before).

src/apify_client/clients/base/base_client.py
```diff
@@ -0,0 +1,37 @@
...
+        if self.resource_id is not None:
+            self.safe_id = _to_safe_id(self.resource_id)
+            self.url = f'{self.url}/{self.safe_id}'
+
+    def _url(self, path: str = None) -> str:
+        if path is not None:
+            return f'{self.url}/{path}'
+        return self.url
```
mhamas, line 32: Can use ternary if
```
return f'{self.url}/{path}' if path is not None else self.url
```

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -0,0 +1,227 @@
...
+        *,
+        offset: int = None,
+        limit: int = None,
+        clean: bool = None,
+        desc: bool = None,
+        fields: List[str] = None,
+        omit: List[str] = None,
+        unwind: str = None,
+        skip_empty: bool = None,
+        skip_hidden: bool = None,
```
mhamas, line 53: All of these should have type `Optional[...]` since they can be `Non`

src/apify_client/clients/resource_clients/dataset.py
mhamas: +1 for Ondra. I think we could actually have internal function `_download_items` which returns different types but then create thin wrapper for stream and buffer versions.

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -0,0 +1,227 @@
...
+            'count': len(data),  # because x-apify-pagination-count returns invalid values when hidden/empty items are skipped
+            'limit': int(response.headers['x-apify-pagination-limit']),  # API returns 999999999999 when no limit is used
+        }
+
+    def download_items(
+        self,
+        download_format: str = 'json',
+        *,
+        as_file: bool = False,
+        offset: int = None,
```
mhamas, line 117: again `Optional[]`

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -0,0 +1,227 @@
...
+            parse_response=False,
+        )
+
+        if as_file:
+            response.raw.decode_content = True
+            return response.raw
+        else:
+            return response.content
+
+    def push_items(self, items: Union[str, Dict, List[str], List[Dict]]) -> None:
```
mhamas, line 204: cannot `items` be `List` of something else than `Dict` and `str`?

src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -0,0 +1,152 @@
...
+        return self._update(new_fields)
+
+    def delete(self) -> None:
+        """Deletes the key-value store.
+
+        https://docs.apify.com/api/v2#/reference/key-value-stores/store-object/delete-store
+        """
+        return self._delete()
+
+    def list_keys(self, *, limit: int = None, exclusive_start_key: str = None) -> Any:
```
mhamas, line 46: `Optional[int]` etc

src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -0,0 +1,152 @@
...
+
+        response = self.http_client.call(
+            url=self._url('keys'),
+            method='GET',
+            params=request_params,
+        )
+
+        return _parse_date_fields(_pluck_data(response.json()))
+
+    def get_record(self, key: str, *, as_bytes: bool = False, as_file: bool = False) -> Optional[Dict]:
```
mhamas, line 68: what about creating `Enum` for type of response instead of two boolean flags? I think would be cleaner that way

src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -0,0 +1,152 @@
...
+        https://docs.apify.com/api/v2#/reference/key-value-stores/record/get-record
+
+        Args:
+            key (str): Key of the record to retrieve
+            as_bytes (bool, optional): Whether to retrieve the record as unparsed bytes, default False
+            as_file (bool, optional): Whether to retrieve the record as a file-like object, default False
+        """
+        try:
+            if as_bytes and as_file:
+                raise ValueError('You cannot have both as_bytes and as_file set.')
```
mhamas, line 80: if we pass enum we can get rid of this

src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -0,0 +1,152 @@
...
+        """Sets a value to the given record in the key-value store.
+
+        https://docs.apify.com/api/v2#/reference/key-value-stores/record/put-record
+
+        Args:
+            key (str): The key of the record to save the value to
+            value (Any): The value to save into the record
+            content_type (str, optional): The content type of the saved value
+        """
+        # TODO revisit this when it's decided whether we keep using the signed URL route or not
```
mhamas, line 111: afaik it's been decided that we will move away from it

src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -0,0 +1,152 @@
...
+        if not content_type:
+            if _is_file_or_bytes(value):
+                content_type = 'application/octet-stream'
+            elif isinstance(value, str):
+                content_type = 'text/plain; charset=utf-8'
+            else:
+                content_type = 'application/json; charset=utf-8'
+
+        if 'application/json' in content_type and not _is_file_or_bytes(value) and not isinstance(value, str):
+            value = json.dumps(value, indent=2)
```
mhamas, line 124: using `indent=...` actually makes JSON larger. Since we are sending this over network, let's do it without it unless there is good reason to keep it.

src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        self.http_client.call(
+            url=self._url(f'records/{key}'),
+            method='DELETE',
+            params=self._params(),
+        )
+
+
+def _is_file_or_bytes(value: Any) -> bool:
+    return isinstance(value, (bytes, bytearray, io.IOBase))
```
mhamas, line 152: what about moving this to some utils file? It's not specific to this subclient.

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -0,0 +1,152 @@
...
+class RequestQueueClient(ResourceClient):
+    """Sub-client for manipulating a single request queue."""
+    def __init__(self, *args: Any, client_key: str = None, **kwargs: Any) -> None:
+        """Initializes the RequestQueueClient.
+
+        Args:
+            client_key (str): A unique identifier of the client accessing the request queue
+        """
+        super().__init__(*args, resource_path='request-queues', **kwargs)
+        self.client_key = client_key
```
mhamas, line 17: Why the client key is used in `request queue client` but not e.g. in `key value store client`?

src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -0,0 +1,152 @@
...
+from typing import Any, Dict, Optional
+
+from ..._errors import ApifyApiError
+from ..._utils import _catch_not_found_or_throw, _parse_date_fields, _pluck_data
+from ..base.resource_client import ResourceClient
+
+
+class KeyValueStoreClient(ResourceClient):
+    """Sub-client for manipulating a single key-value store."""
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
mhamas, line 12: Have you pls verified that when you are creating this client from other code that it gives you correct argument hints?

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -0,0 +1,152 @@
...
+    def delete_request(self, request_id: str) -> None:
+        """Deletes a request from the queue.
+
+        https://docs.apify.com/api/v2#/reference/request-queues/request/delete-request
+
+        Args:
+            request_id (str): ID of the request to delete.
+        """
+        request_params = self._params(
+            clientKey=self.client_key,
```
mhamas, line 145: what is this key exactly?

tests/unit/test_utils.py
```diff
@@ -0,0 +1,102 @@
...
+    attempt_counter = 0
+    with pytest.raises(RetryableException):
+        _retry_with_exp_backoff(returns_on_fifth_attempt, max_retries=4, backoff_base_millis=1)
+    assert attempt_counter == 4
+
+    # Bails when the bail function is called
+    attempt_counter = 0
+    with pytest.raises(BailException):
+        _retry_with_exp_backoff(bails_on_third_attempt, backoff_base_millis=1)
+    assert attempt_counter == 3
```
mhamas, line 102: Let's use native python `unittest` framework for this. When you create a class with unit tests which subclasses `unittest.TestCase` you get lot of fancy comparison methods such as `self.assertEqual()` that ar easier to read than plain `assert X == Y`

```
import time
import unittest
from datetime import datetime, timezone
from typing import Any, Callable

from apify_client._utils import _parse_date_fields, _pluck_data, _retry_with_exp_backoff, _to_safe_id


class UtilsTest(unittest.TestCase):

    def test__to_safe_id(self) -> None:
        self.assertEqual(_to_safe_id('abc'), 'abc')
        self.assertEqual(_to_safe_id('abc/def'), 'abc~def')
        self.assertEqual(_to_safe_id('abc/def'), 'abc~def')

    def test__parse_date_fields(self) -> None:
        # works correctly on empty dicts
        self.assertEqual(_parse_date_fields({}), {})

        # correctly parses dates on fields ending with -At
        expected_datetime = datetime(2016, 11, 14, 11, 10, 52, 425000, timezone.utc)
        self.assertEqual(
            _parse_date_fields({'createdAt': '2016-11-14T11:10:52.425Z'}),
            {'createdAt': expected_datetime}
        )

        # doesn't parse dates on fields not ending with -At
        self.assertEqual(
            _parse_date_fields({'saveUntil': '2016-11-14T11:10:52.425Z'}),
            {'saveUntil': '2016-11-14T11:10:52.425Z'}
        )

        # parses dates in dicts in lists
        expected_datetime = datetime(2016, 11, 14, 11, 10, 52, 425000, timezone.utc)
        self.assertEqual(
            _parse_date_fields([{'createdAt': '2016-11-14T11:10:52.425Z'}]),
            [{'createdAt': expected_datetime}]
        )

        # parses nested dates
        expected_datetime = datetime(2020, 2, 29, 10, 9, 8, 100000, timezone.utc)
        self.assertEqual(
            _parse_date_fields({'a': {'b': {'c': {'createdAt': '2020-02-29T10:09:08.100Z'}}}}),
            {'a': {'b': {'c': {'createdAt': expected_datetime}}}}
        )

        # doesn't parse dates nested too deep
        expected_datetime = datetime(2020, 2, 29, 10, 9, 8, 100000, timezone.utc)
        self.assertEqual(
            _parse_date_fields({'a': {'b': {'c': {'d': {'createdAt': '2020-02-29T10:09:08.100Z'}}}}}),
            {'a': {'b': {'c': {'d': {'createdAt': '2020-02-29T10:09:08.100Z'}}}}}
        )

        # doesn't die when the date can't be parsed
        self.assertEqual(
            _parse_date_fields({'createdAt': 'NOT_A_DATE'}),
            {'createdAt': 'NOT_A_DATE'}
        )

    def test__pluck_data(self) -> None:
        # works correctly when data is present
        self.assertEqual(_pluck_data({'data': {}}), {})
        self.assertEqual(_pluck_data({'a': 'b', 'data': {'b': 'c'}}), {'b': 'c'})

        # throws the right error when it is not
        with self.assertRaises(ValueError):
            _pluck_data({'a': 'b'})
        with self.assertRaises(ValueError):
            _pluck_data(None)
        with self.assertRaises(ValueError):
            _pluck_data('{"a": "b"}')

    def test__retry_with_exp_backoff(self) -> None:
        attempt_counter = 0

        class RetryableException(Exception):
            pass

        class BailException(Exception):
            pass

        def returns_on_fifth_attempt(bail: Callable, attempt: int) -> Any:
            nonlocal attempt_counter
            attempt_counter += 1

            if attempt == 5:
                return 'SUCCESS'
            raise RetryableException()

        def bails_on_third_attempt(bail: Callable, attempt: int) -> Any:
            nonlocal attempt_counter
            attempt_counter += 1

            if attempt == 3:
                bail(BailException())
            raise RetryableException()

        # Returns the correct result after the correct time (should take 100 + 200 + 400 + 800 = 1500 ms)
        start = time.time()
        result = _retry_with_exp_backoff(returns_on_fifth_attempt, backoff_base_millis=100, backoff_factor=2, random_factor=0)
        elapsed_time_seconds = time.time() - start
        self.assertEqual(result, 'SUCCESS')
        self.assertEqual(attempt_counter, 5)
        self.assertGreater(elapsed_time_seconds, 1.4)
        self.assertLess(elapsed_time_seconds, 2.0)

        # Stops retrying when failed for max_retries times
        attempt_counter = 0
        with self.assertRaises(RetryableException):
            _retry_with_exp_backoff(returns_on_fifth_attempt, max_retries=4, backoff_base_millis=1)
        self.assertEqual(attempt_counter, 4)

        # Bails when the bail function is called
        attempt_counter = 0
        with self.assertRaises(BailException):
            _retry_with_exp_backoff(bails_on_third_attempt, backoff_base_millis=1)
        self.assertEqual(attempt_counter, 3)

```

mtrunkat: To me, as somebody who doesn't know python at all, this syntax

`assert _to_safe_id('abc~def') == 'abc~def'`

is easier to read than

`self.assertEqual(_to_safe_id('abc/def'), 'abc~def')`

but that is not the right metric. But I remember that guys on the call were mentioning pytest as their tool of choice, isn't it?

mhamas: `pytest` is the thing which runs these tests under the hood (they support this paradigm).

The advantage of `self.assertXY` is that it's much more robust than `assert X == Y`. You can do all kinds of asserts, you can expect exceptions etc.

List of asserts is here: https://docs.python.org/3/library/unittest.html#assert-methods
You can even check that lists have same elements regardless of order with `assertCountEqual` (this is super useful when comparing lists that results as taking some keys / values from dictionaries :))

mtrunkat (approved):

src/apify_client/clients/resource_clients/dataset.py
fnesveda: Updated this based on our previous discussion with Ondra to add the function `iterate_items`, which iterates over the dataset items, downloading them as necessary.

src/apify_client/_http_client.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import Any, Callable, Dict, List, Optional, Union
+
+import requests
+from requests.exceptions import ConnectionError, Timeout
+
+from ._errors import ApifyApiError, InvalidResponseBodyError
+from ._utils import _retry_with_exp_backoff
+
+NO_CONTENT_STATUS_CODE = 204
+RATE_LIMIT_EXCEEDED_STATUS_CODE = 429
```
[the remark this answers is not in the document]
fnesveda: I used the enum from the `http` module for this.

src/apify_client/_http_client.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def call(
+        self,
+        *,
+        method: str,
+        url: str,
+        headers: Dict = None,
+        params: Dict = None,
+        data: str = None,
+        json: Union[str, Dict, List[str], List[Dict]] = None,
```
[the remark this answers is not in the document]
fnesveda: I added a type `JSONSerializable` which should cover all the possible supported values.

src/apify_client/_http_client.py
```diff
@@ -0,0 +1,108 @@
...
+                        _maybe_parsed_body = response.raw
+                    else:
+                        _maybe_parsed_body = response.content
+                    setattr(response, '_maybe_parsed_body', _maybe_parsed_body)
+                    return response
+
+            except (ConnectionError, Timeout, InvalidResponseBodyError) as e:
+                raise e
+            except Exception as e:
+                bail(e)
```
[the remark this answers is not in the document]
fnesveda: It works, `bail` is passed as an argument to the function.

src/apify_client/_http_client.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+    @staticmethod
+    def _maybe_parse_response(response: requests.models.Response) -> Any:
+        if response.status_code == NO_CONTENT_STATUS_CODE:
+            return None
+
+        content_type = ''
+        if 'content-type' in response.headers:
+            content_type = response.headers['content-type'].split(';')[0].strip()
```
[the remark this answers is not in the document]
fnesveda: That's not very readable, I prefer the current way over the Python ternaries.

src/apify_client/_http_client.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @staticmethod
+    def _parse_params(params: Optional[Dict]) -> Optional[Dict]:
+        if params is None:
+            return None
+
+        parsed_params = {}
+        for key, value in params.items():
+            if isinstance(value, bool):
+                parsed_params[key] = int(value)
```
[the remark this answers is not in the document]
fnesveda: It's because otherwise the boolean would be formatted to the request URL as a string, which our API does not support. I added a comment.

src/apify_client/_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    func: Callable[[BailType, int], T],
```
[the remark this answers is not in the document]
fnesveda: This function is only used internally in one place, I agree it's not very flexible but I think it's good enough for now. We can rewrite it later, let's keep it like it is now so we can finally merge this, and rewrite it later.

src/apify_client/clients/base/actor_job_base_client.py
fnesveda: Sort of, it will be used when implementing this client, so I just left it there commented out instead of deleting it.

src/apify_client/clients/base/base_client.py
fnesveda: Again, I find the current way a bit more readable than the Python ternaries, I'd keep it as it is.

src/apify_client/clients/resource_clients/key_value_store.py
fnesveda: Yeah, it's not ideal, we want to update the whole get_record logic in both the JS and Python clients and maybe split it into multiple functions. I'd go for `get_record`, `download_record` and `stream_record`, but the JS client has to have it somehow backwards compatible, so we still need to figure this out. I'll leave it like it is for now and revisit it later.

src/apify_client/clients/resource_clients/key_value_store.py
fnesveda: I added it so that the behavior between the JS and Python client is the same, it improves readability when displayed in the app and if we `gzip` the POST requests (which we're currently not doing in this client, but it's on the roadmap), the size difference should not be too big. But I'm not hell-bent on keeping it, what do you think, @mnmkng

src/apify_client/clients/resource_clients/key_value_store.py
fnesveda: Yup, it works for me in Visual Studio Code.

tests/unit/test_utils.py
fnesveda: Updated it to use `self.assertEqual`, it is indeed better, it displays errors better etc.

src/apify_client/clients/resource_clients/request_queue.py
fnesveda: "Client" here does not mean the API client specifically, but anything accessing the request queue, it's used so that the request queue can differentiate between different applications accessing it in parallel, so each application (client) uses it's own key for that.

src/apify_client/clients/resource_clients/request_queue.py
fnesveda: See comment above

fnesveda:
Thanks for the thorough review, @mhamas! I've addressed your comments, I agreed with most things and updated the code accordingly, the rest we already talked about on Friday. Can you please take another look?

README.md
```diff
@@ -3,4 +3,42 @@
 This will be an official client for [Apify API](https://www.apify.com/docs/api/v2).
 It's still work in progress, so please don't use it yet!
 
+TODO: write a proper README
```
mhamas, line 6: maybe let's remove this TODO since everybody will see on the first page - not good UX

README.md
```diff
@@ -3,4 +3,42 @@
...
 
+## Installation
+
+Requires Python 3.7+
```
mhamas, line 10: out of interest, is the client expected to work with 3.8 as well?

lint_and_test.sh
```diff
@@ -0,0 +1,40 @@
...
+
+set -u
+
+checks_to_run=${1:-"everything"}
+shift || true
+
+
+lint() {
+    echo "⛄️ flake8 check..."
+    python3 -m flake8 src tests
```
mhamas, line 11: what about `isort` and `autopep` tests?

src/apify_client/_types.py
```diff
@@ -0,0 +1,6 @@
+from typing import Any, Dict, List, Union
+
+# Type for representing json-serializable values
+# It's close enough to the real thing supported by json.parse, and the best we can do until mypy supports recursive types
+# It was suggested in a discussion with (and approved by) Guido van Rossum, so I'd consider it correct enough
+JSONSerializable = Union[str, int, float, bool, None, Dict[str, Any], List[Any]]
```
mhamas, line 6: sweet :-)

src/apify_client/clients/resource_clients/key_value_store.py
mhamas: ok, please add TODO

src/apify_client/clients/resource_clients/key_value_store.py
mhamas: Please, at least add big TODO here. From my previous experience the difference was couple of X for large JSONs. Thx

mhamas (approved):
Thx @fnesveda

lint_and_test.sh
fnesveda: `flake8` checks PEP8 compliance, so it should lint according to the same rules as `autopep` is formatting, and I added the `flake8-isort` plugin instead of using `isort` directly, it's easier to integrate with VSCode that way.

README.md
fnesveda: Good point, I'll remove the TODO for now, but we should still write a proper README once the code is nearly done, with some usage examples etc.

README.md
fnesveda: Yup, it should be forward-compatible, it would be pretty weird if it wasn't.

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -0,0 +1,377 @@
...
+            'total': int(response.headers['x-apify-pagination-total']),
+            'offset': int(response.headers['x-apify-pagination-offset']),
+            'count': len(data),  # because x-apify-pagination-count returns invalid values when hidden/empty items are skipped
+            'limit': int(response.headers['x-apify-pagination-limit']),  # API returns 999999999999 when no limit is used
+        }
+
+    def iterate_items(
+        self,
+        offset: int = 0,
+        limit: int = 99999999999,
```
Equidem, line 118: I don't like this, but I guess it is an API issue, not the client. Also, there is one 9 less than in the comment a few lines above this ("API returns 999999999999 when no limit is used").

src/apify_client/clients/resource_clients/dataset_collection.py
```diff
@@ -0,0 +1,40 @@
...
+            offset (int, optional): What dataset to include as first when retrieving the list
+            desc (bool, optional): Whether to sort the datasets in descending order based on their modification date
+
+        Returns:
+            The list of available datasets matching the specified filters.
+        """
+        return self._list(unnamed=unnamed, limit=limit, offset=offset, desc=desc)
+
+    def get_or_create(self, *, name: str = '') -> Dict:
+        """Retrieve a named dataset, or creates a new one when it doesn't exist.
```
Equidem, line 30: Should be "create", not "creates".

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+class DatasetClient(ResourceClient):
+    """Sub-client for manipulating a single dataset."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the DatasetClient."""
+        super().__init__(*args, resource_path='datasets', **kwargs)
+
+    def get(self) -> Optional[Dict]:
```
Equidem, line 15: (applies to all the other clients like this too)
Maybe we could override the [] operator and allow extracting these values directly using the DatasetClient instance? It would be a little more user friendly and wouldn't require much additional work. We could also easily do the same for the setter use-case of [].

Equidem (approved):
Nice work! Most of the stuff I would object to has already been mentioned, so I only have a few nitpicks.

src/apify_client/clients/resource_clients/dataset.py
fnesveda: That's actually not the same thing, it is a client problem not directly related to the API and to the comment above, it was just supposed to be a really big number used as a default, big enough limit, so that you can iterate through all the items. I don't like this either, so I rewrote the logic so it's cleaner.

src/apify_client/clients/resource_clients/dataset_collection.py
fnesveda: Good catch!

src/apify_client/clients/resource_clients/dataset.py
fnesveda: I think this would be wrong conceptually - when you call `client.dataset('someDatasetId')`, you're getting another client, not the dataset itself. Implementing `[]` would mix these two concepts together, and I think in the end it would be just more confusing to users.

Also, we'd have to download the whole resource again and again for every property we would be accessing. I think that would be weird as well, when you call a function like `.get()`, you kinda intuitively expect that it might take some time, but when you access something via the `[]` operator, you intuitively expect it to be already there and accessible quickly.

We can revisit this later, but for this first version I'd skip it and stick just to the functions.

PR #24, opened 2021-01-13
Add user client

mhamas:
This PR adds a user client with a single `get` method for retrieving user data.

**Usage**
```

client = ApifyClient(token="YOUR_TOKEN")
print(client.user(user_id="some_user_id").get())
```


lint_and_test.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -u
+set -ue
```
mhamas, line 3: `-e` needs to be there so that the bash script errors out if something (such as lint) fails

fnesveda: But now if for example you run this with `everything`, and lint fails, it will not run type checks at all, right? It would be better if we ran all the tests and then errored out if any one of the checks failed, so you would get the most information from that one run, and wouldn't have to run it multiple times.

fnesveda (changes requested):
Nice! Just a small comment about the erroring out of `lint_and_test.sh`, but I'm happy this is starting to move somewhere :)

mhamas:
With `-e` it stops whenever it encounters the first error. So if lint fails, the script stops at that point. This is a standard way to do it that I've seen before. And for me, that's the behavior I want since I don't want to wait for unit tests to finish if there are some linting or typing errors (which might potentially affect the tests as well). I rather rerun the script many times. Also, the errors are much easier to spot since the last line is your error. 

Also, without `-e`, scripts exits with `0` even if `flake8` fails. Which we don't want, since we want to use this on CI.

fnesveda:
@mhamas OK, makes sense, let's keep it this way then :) I'll merge this right after the big PR is merged, once I get a review from Matěj Sochor.

fnesveda (approved):

PR #25, opened 2021-01-14
Add client for manipulating a single build object

mhamas:
This PR adds a sub-client for manipulating a single actor build, mimicking the methods that exist on JS client.

### DIFF
--- a/src/apify_client/clients/base/actor_job_base_client.py
+++ b/src/apify_client/clients/base/actor_job_base_client.py
@@ -20,7 +20,7 @@ class ActorJobBaseClient(ResourceClient):
         job: Optional[Dict] = None
         seconds_elapsed = 0
 
-        while(should_repeat):
+        while should_repeat:
             wait_for_finish = DEFAULT_WAIT_FOR_FINISH
             if wait_secs is not None:
                 wait_for_finish = wait_secs - seconds_elapsed
@@ -43,11 +43,11 @@ class ActorJobBaseClient(ResourceClient):
                     # Early return here so that we avoid the sleep below if not needed
                     return job
 
-                # It might take some time for database replicas to get up-to-date so sleep a bit before retrying
-                time.sleep(250)
-
             except ApifyApiError as exc:
                 _catch_not_found_or_throw(exc)
+
+            # It might take some time for database replicas to get up-to-date so sleep a bit before retrying
+            time.sleep(250)
 
         return job
 
--- a/src/apify_client/clients/base/actor_job_base_client.py
+++ b/src/apify_client/clients/base/actor_job_base_client.py
@@ -1,40 +1,60 @@
 import math
+import time
 from datetime import datetime
 from typing import Dict, Optional
 
 from ..._consts import TERMINAL_ACTOR_JOB_STATUSES, ActorJobStatus
-from ..._utils import _parse_date_fields, _pluck_data
+from ..._errors import ApifyApiError
+from ..._utils import _catch_not_found_or_throw, _parse_date_fields, _pluck_data
 from .resource_client import ResourceClient
+
+DEFAULT_WAIT_FOR_FINISH = 999999
 
 
 class ActorJobBaseClient(ResourceClient):
     """Base sub-client class for actor runs and actor builds."""
 
-    def _wait_for_finish(self, wait_secs: Optional[int] = None) -> Dict:
+    def _wait_for_finish(self, wait_secs: Optional[int] = None) -> Optional[Dict]:
         started_at = datetime.now()
         should_repeat = True
         job: Optional[Dict] = None
         seconds_elapsed = 0
 
         while(should_repeat):
-            wait_for_finish = 9999
+            wait_for_finish = DEFAULT_WAIT_FOR_FINISH
             if wait_secs is not None:
                 wait_for_finish = wait_secs - seconds_elapsed
 
-            response = self.http_client.call(
-                url=self.url,
-                method='GET',
-                params=self._params(wait_for_finish=wait_for_finish),
-            )
-            job = _parse_date_fields(_pluck_data(response.json()))
+            try:
+                response = self.http_client.call(
+                    url=self._url(),
+                    method='GET',
+                    params=self._params(waitForFinish=wait_for_finish),
+                )
+                job = _parse_date_fields(_pluck_data(response.json()))
 
-            seconds_elapsed = math.floor(((datetime.now() - started_at).total_seconds()))
-            if (
-                ActorJobStatus(job['status']) in TERMINAL_ACTOR_JOB_STATUSES or (wait_secs is not None and seconds_elapsed >= wait_secs)
-            ):
-                should_repeat = False
+                seconds_elapsed = math.floor(((datetime.now() - started_at).total_seconds()))
+                if (
+                    ActorJobStatus(job['status']) in TERMINAL_ACTOR_JOB_STATUSES or (wait_secs is not None and seconds_elapsed >= wait_secs)
+                ):
+                    should_repeat = False
 
-        if job is None:
-            raise ValueError(f"Waiting for actor job with ID {self.resource_id} failed.")
+                if not should_repeat:
+                    # Early return here so that we avoid the sleep below if not needed
+                    return job
+
+                # It might take some time for database replicas to get up-to-date so sleep a bit before retrying
+                time.sleep(250)
+
+            except ApifyApiError as exc:
+                _catch_not_found_or_throw(exc)
 
         return job
+
+    def _abort(self) -> Dict:
+        response = self.http_client.call(
+            url=self._url('abort'),
+            method='POST',
+            params=self._params(),
+        )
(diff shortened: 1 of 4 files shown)


src/apify_client/clients/resource_clients/build.py
```diff
@@ -0,0 +1,46 @@
...
+
+        Returns:
+            The data of the aborted actor build
+        """
+        response = self.http_client.call(
+            url=self._url('abort'),
+            method='POST',
+            params=self._params(),
+        )
+        return _parse_date_fields(_pluck_data(response.json()))
```
fnesveda, line 37: We could actually move this implementation into `ActorJobBaseClient._abort()`, since it will also be used in the run client in exactly the same way.

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -1,13 +1,40 @@
...
+
+        while(should_repeat):
+            wait_for_finish = 9999
+            if wait_secs is not None:
+                wait_for_finish = wait_secs - seconds_elapsed
+
+            response = self.http_client.call(
+                url=self.url,
+                method='GET',
+                params=self._params(wait_for_finish=wait_for_finish),
```
fnesveda, line 27: You have to have `waitForFinish=wait_for_finish` here, the API URL parameter is camel-cased, see https://docs.apify.com/api/v2#/reference/actor-builds/build-object/get-build.

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -1,13 +1,40 @@
...
+        job: Optional[Dict] = None
+        seconds_elapsed = 0
+
+        while(should_repeat):
+            wait_for_finish = 9999
+            if wait_secs is not None:
+                wait_for_finish = wait_secs - seconds_elapsed
+
+            response = self.http_client.call(
+                url=self.url,
```
fnesveda, line 25: Let's use `url=self._url()` here for consistency with other resource clients, if we ever decide to change the URL generating logic we can change it in just that one function.

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -1,13 +1,40 @@
...
-        # TODO implement when implementing run and build endpoints
-        pass
+    def _wait_for_finish(self, wait_secs: Optional[int] = None) -> Dict:
+        started_at = datetime.now()
+        should_repeat = True
+        job: Optional[Dict] = None
+        seconds_elapsed = 0
+
+        while(should_repeat):
+            wait_for_finish = 9999
```
fnesveda, line 20: We could have the `wait_secs` parameter have 9999 as a default value, and not have it optional - that way you can simplify the logic here, and it will be explicitly obvious to the caller how long will this function wait by default. Also, let's use 999999 just as the JS client, sometimes you might want to wait for more than ~3 hours for the run to finish.

src/apify_client/clients/resource_clients/build.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def wait_for_finish(self, *, wait_secs: Optional[int] = None) -> Dict:
```
fnesveda, line 39: Could you add documentation for the wait_secs param as well? I'm surprised D417 from `pydocstyle` didn't catch this, it should be enabled in our `flake8` config. I'll take a look.

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -1,13 +1,40 @@
...
+            )
+            job = _parse_date_fields(_pluck_data(response.json()))
+
+            seconds_elapsed = math.floor(((datetime.now() - started_at).total_seconds()))
+            if (
+                ActorJobStatus(job['status']) in TERMINAL_ACTOR_JOB_STATUSES or (wait_secs is not None and seconds_elapsed >= wait_secs)
+            ):
+                should_repeat = False
+
+        if job is None:
```
fnesveda, line 37: This will never happen, if the job is missing it will throw an `ApifyApiError` from the `http_client` call. You should wrap that call with a try - catch simiar to the one in `ResourceClient._get()`.

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -1,13 +1,40 @@
...
+            response = self.http_client.call(
+                url=self.url,
+                method='GET',
+                params=self._params(wait_for_finish=wait_for_finish),
+            )
+            job = _parse_date_fields(_pluck_data(response.json()))
+
+            seconds_elapsed = math.floor(((datetime.now() - started_at).total_seconds()))
+            if (
+                ActorJobStatus(job['status']) in TERMINAL_ACTOR_JOB_STATUSES or (wait_secs is not None and seconds_elapsed >= wait_secs)
```
fnesveda, line 33: If `job` would be None this would throw because it could not call `__getitem__()`.

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -1,13 +1,40 @@
...
+            ):
+                should_repeat = False
+
```
fnesveda, line 36: We should add some delay between the retries here, like it's done in the JS client, in case you try to get a job object which was not propagated across all the MongoDB replicas yet.

fnesveda (changes requested):

Equidem (commented):
I just went through the code and didn't notice anything note-worthy other than what @fnesveda already mentioned.

src/apify_client/clients/base/actor_job_base_client.py
mhamas: This won't happen as `_pluck_data` always returns Dict (or throws). I actually had the check there but `mypy` complained it is redundant.

src/apify_client/clients/base/actor_job_base_client.py
mhamas: I'm changing it, but I don't think this is the case. It worked with `wait_for_finish` as well, I guess the HTTP client must be translating this in payload.

src/apify_client/clients/base/actor_job_base_client.py
mhamas: It'd make the logic simpler, but I think that it's cleaner to pass `None` when user means "I don't care about wait secs. I just want to wait indefinitely". So I'm for keeping optional here. Changing default to 999999

src/apify_client/clients/base/actor_job_base_client.py
```diff
@@ -1,13 +1,60 @@
...
+                    should_repeat = False
+
+                if not should_repeat:
+                    # Early return here so that we avoid the sleep below if not needed
+                    return job
+
+                # It might take some time for database replicas to get up-to-date so sleep a bit before retrying
+                time.sleep(250)
+
+            except ApifyApiError as exc:
```
fnesveda, line 49: This catch should be before the `time.sleep()`, otherwise if the job is not found it will never sleep and just retry the request immediately.

The point here is that when you do `client.actor('actor_id').call(...)`, it will start the actor, wait for it to finish and then return the run object once it's finished. It can happen, though, that when you start the actor and immediately try to access the run object, the newly created run object might not be propagated to all the DB replicas, and the first API request to get that run object might not find it, so we should wait a bit before trying again to give the DB replicas a chance to sync all the changes.

Perhaps we should also improve the `should_repeat` condition to not wait for the DB replica sync for too long, otherwise this might never return even if the run object really doesn't exist.

fnesveda (changes requested):
~~It's a shame that you can't reply to resolved conversations on GitHub, wasn't that possible until recently?~~ Edit: now it works, but I would swear it didn't 1 hour ago.

src/apify_client/clients/base/actor_job_base_client.py
fnesveda: Just tested it with locally running API and the snake_case params do not get converted to camelCase params, the `wait_for_finish` param just did not get parsed, so dunno what went on there.

src/apify_client/clients/base/actor_job_base_client.py
fnesveda: Agreed, it makes sense now that internally we are waiting 999999 seconds, which is basically infinitely.

src/apify_client/clients/base/actor_job_base_client.py
mhamas: I moved the `sleep` but I'm not sure what you mean by modifying `should_repeat` condition. thx

fnesveda: I meant that right now, when you would do `wait_for_finish` on a non-existing actor job, you would be here catching `not found` errors in a loop, stuck forever, instead of just returning `None` if it really doesn't exist. We should allow for an early exit from this loop if there was enough time for the DB replicas to sync and the actor job is still `None`.
I think maybe 1 second should be enough for the replicas to sync? So if the API still returns that the job does not exist even after that time, we should just return None, as the job really does not exist.

fnesveda (changes requested):

fnesveda (approved):

PR #26, opened 2021-01-15
Add subclient for querying all builds for a user

mhamas:
Note that this client queries for builds of **a user**, not a particular actor.

### DIFF
--- a/src/apify_client/clients/resource_clients/build_collection.py
+++ b/src/apify_client/clients/resource_clients/build_collection.py
@@ -1,0 +1,24 @@
+from typing import Any, Dict, Optional
+
+from ..base.resource_collection_client import ResourceCollectionClient
+
+
+class BuildCollectionClient(ResourceCollectionClient):
+    """Sub-client for listing user builds.
+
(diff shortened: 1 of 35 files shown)


src/apify_client/clients/resource_clients/build_collection.py
```diff
@@ -0,0 +1,24 @@
...
+    """Sub-client for listing user builds.
+
+    Note that this client is not specific for a particular actor but queries all builds for a user based on the provided API token.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the BuildCollectionClient."""
+        super().__init__(*args, resource_path='actor-builds', **kwargs)
+
+    def list(self, *, limit: Optional[int] = None, offset: Optional[int] = None, desc: Optional[bool] = None) -> Dict:
```
fnesveda, line 16: Could you document the args as well? D417 didn't fire here again for some reason.

fnesveda (changes requested):

Equidem (approved):

fnesveda (approved):

PR #27, opened 2021-01-18
Fix parsing errors from S3

fnesveda:
When attempting to retrieve a non-existing record from a key-value store, we get an error directly from S3, which is XML-formatted, instead of from our API servers, which return it in JSON. This fixes the parsing of those errors, enabling correct behaviour (returning None) for those missing records.

I've also set the default `Accept` header to `application/json, */*`, in order to prefer JSON responses and maintain compatibility with the JS client.

### DIFF
--- a/src/apify_client/_errors.py
+++ b/src/apify_client/_errors.py
@@ -23,17 +23,19 @@ class ApifyApiError(ApifyClientError):
 
         Args:
             response: The response to the failed API call
-            attempt: Which retry was the request that failed
+            attempt: Which attempt was the request that failed
         """
         self.message: Optional[str] = None
         self.type: Optional[str] = None
 
-        response_data = response.json()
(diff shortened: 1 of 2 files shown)


Equidem (commented):
It might be better to detect the specific error we get from S3 instead of any non-json error, which is what this does if I am correct. We could lose information about important problems this way later on.

fnesveda:
@Equidem This just fixes the parsing of the errors which we get from S3 into the `ApifyApiError.message` field, the actual check for whether it's a "not found" error is in https://github.com/apify/apify-client-python/blob/master/src/apify_client/_utils.py#L120, and that checks for both options, if the error was JSON or if it was not. No information is actually lost here.

Equidem (approved):
I misunderstood, sorry. All good then.

mhamas (approved):

PR #29, opened 2021-02-05
Schedule subclient

mhamas:
Subclient for manipulating single schedule or schedule collection.

Note that I had to get a bit creative in `create` function in `ScheduleCollectionClient`. The arguments listed in API docs (https://docs.apify.com/api/v2#/reference/schedules/schedules-collection/create-schedule) are not 100% accurate. `cronExpression`, `isEnabled` and `isExclusive` seem to be required and also it seems it doesn't matter what `userId` you pass, it'll be set based on token.

### DIFF
--- a/src/apify_client/clients/resource_clients/schedule.py
+++ b/src/apify_client/clients/resource_clients/schedule.py
@@ -6,14 +6,14 @@ from ..base.resource_client import ResourceClient
 
 
 class ScheduleClient(ResourceClient):
-    """Sub-client for manipulating schedules."""
+    """Sub-client for manipulating a single schedule."""
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         """Initialize the ScheduleClient."""
         super().__init__(*args, resource_path='schedules', **kwargs)
 
     def get(self) -> Optional[Dict]:
-        """Return information about schedule.
+        """Return information about the schedule.
 
         https://docs.apify.com/api/v2#/reference/schedules/schedule-object/get-schedule
 
@@ -32,7 +32,7 @@ class ScheduleClient(ResourceClient):
         actions: Optional[List[Dict]] = None,
         description: Optional[str] = None,
         timezone: Optional[str] = None,
-    ) -> Optional[Dict]:
+    ) -> Dict:
         """Update the schedule with specified fields.
 
         https://docs.apify.com/api/v2#/reference/schedules/schedule-object/update-schedule
--- a/src/apify_client/clients/resource_clients/schedule.py
+++ b/src/apify_client/clients/resource_clients/schedule.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict, List, Optional
 
 from ..._errors import ApifyApiError
-from ..._utils import _catch_not_found_or_throw, _parse_date_fields, _pluck_data
+from ..._utils import _catch_not_found_or_throw, _pluck_data_as_list
 from ..base.resource_client import ResourceClient
 
 
@@ -56,7 +56,7 @@ class ScheduleClient(ResourceClient):
                 method='GET',
                 params=self._params(),
             )
-            return _parse_date_fields(_pluck_data(response.json()))
+            return _pluck_data_as_list(response.json())
         except ApifyApiError as exc:
             _catch_not_found_or_throw(exc)
 
--- a/src/apify_client/clients/resource_clients/schedule.py
+++ b/src/apify_client/clients/resource_clients/schedule.py
@@ -43,7 +43,7 @@ class ScheduleClient(ResourceClient):
             is_exclusive: When set to true, don't start actor or actor task if it's still running from the previous schedule.
             name: The name of the schedule to create.
             actions: Actors or tasks that should be run on this schedule. See the API documentation for exact structure.
-            description: Description of this scheedule
+            description: Description of this schedule
             timezone: Timezone in which your cron expression runs (TZ database name from https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)
 
         Returns:
--- a/src/apify_client/clients/resource_clients/schedule.py
+++ b/src/apify_client/clients/resource_clients/schedule.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict, List, Optional
 
 from ..._errors import ApifyApiError
-from ..._utils import _catch_not_found_or_throw, _pluck_data_as_list
+from ..._utils import _catch_not_found_or_throw, _pluck_data_as_list, _snake_case_to_camel_case
 from ..base.resource_client import ResourceClient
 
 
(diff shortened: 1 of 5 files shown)


src/apify_client/clients/resource_clients/schedule.py
```diff
@@ -0,0 +1,63 @@
...
+        """Return information about schedule.
+
+        https://docs.apify.com/api/v2#/reference/schedules/schedule-object/get-schedule
+
+        Returns:
+            The retrieved schedule
+        """
+        return self._get()
+
+    def update(self, new_fields: Dict) -> Optional[Dict]:
```
fnesveda, line 25: It would be nice if this actually accepted the fields to update as named parameters instead of just a dict, as we discussed with Matěj Sochor. Unfortunately, it means duplicating the parameters in code as it stands right now, but since the API should not change too much, it hopefully won't be too much of a problem.

src/apify_client/clients/resource_clients/schedule_collection.py
```diff
@@ -0,0 +1,57 @@
...
+        actions: List[Dict] = [],
+    ) -> Dict:
+        """Create a new schedule.
+
+        https://docs.apify.com/api/v2#/reference/schedules/schedules-collection/create-schedule
+
+        Args:
+            cron_expression: The cron expression used by this schedule
+            is_enabled: True if the schedule should be enabled
+            is_exclusive: True if the schedule is exclusive
```
fnesveda, line 44: This would deserve a better explanation, like "Don't start scheduled actor or task if it is still running from the previous schedule".

src/apify_client/clients/resource_clients/schedule_collection.py
```diff
@@ -0,0 +1,57 @@
...
+        return self._list(limit=limit, offset=offset, desc=desc)
+
+    def create(
+        self,
+        *,
+        cron_expression: str,
+        is_enabled: bool,
+        is_exclusive: bool,
+        name: Optional[str] = None,
+        actions: List[Dict] = [],
```
fnesveda, line 35: You can actually specify a timezone as well (it's not documented in the API docs), it's a string containing the timezone name as specified in the [tz database](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).

fnesveda: You can also specify the schedule description, an optional string.

fnesveda (changes requested):
Wow, I had no idea the API docs for schedules are so bad, we have to fix them. I had just some small comments here.

src/apify_client/clients/resource_clients/schedule.py
```diff
@@ -0,0 +1,83 @@
...
+
+        https://docs.apify.com/api/v2#/reference/schedules/schedule-object/update-schedule
+
+        Args:
+            cron_expression: The cron expression used by this schedule
+            is_enabled: True if the schedule should be enabled
+            is_exclusive: When set to true, don't start actor or actor task if it's still running from the previous schedule.
+            name: The name of the schedule to create.
+            actions: Actors or tasks that should be run on this schedule. See the API documentation for exact structure.
+            description: Description of this scheedule
```
fnesveda, line 46: You have "scheedule" here :)

src/apify_client/clients/resource_clients/schedule_collection.py
```diff
@@ -0,0 +1,63 @@
...
+
+        https://docs.apify.com/api/v2#/reference/schedules/schedules-collection/create-schedule
+
+        Args:
+            cron_expression: The cron expression used by this schedule
+            is_enabled: True if the schedule should be enabled
+            is_exclusive: When set to true, don't start actor or actor task if it's still running from the previous schedule.
+            name: The name of the schedule to create.
+            actions: Actors or tasks that should be run on this schedule. See the API documentation for exact structure.
+            description: Description of this scheedule
```
fnesveda, line 49: Here too

fnesveda (changes requested):
Looking good! Just found one typo.

src/apify_client/clients/resource_clients/schedule.py
```diff
@@ -0,0 +1,83 @@
+from typing import Any, Dict, List, Optional
+
+from ..._errors import ApifyApiError
+from ..._utils import _catch_not_found_or_throw, _pluck_data_as_list, _snake_case_to_camel_case
+from ..base.resource_client import ResourceClient
+
+
+class ScheduleClient(ResourceClient):
+    """Sub-client for manipulating schedules."""
```
Equidem, line 9: "a single schedule"

src/apify_client/clients/resource_clients/schedule.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the ScheduleClient."""
+        super().__init__(*args, resource_path='schedules', **kwargs)
+
+    def get(self) -> Optional[Dict]:
+        """Return information about schedule.
```
Equidem, line 16: "the schedule"

src/apify_client/clients/resource_clients/schedule.py
```diff
@@ -0,0 +1,83 @@
...
+        self,
+        *,
+        cron_expression: Optional[str] = None,
+        is_enabled: Optional[bool] = None,
+        is_exclusive: Optional[bool] = None,
+        name: Optional[str] = None,
+        actions: Optional[List[Dict]] = None,
+        description: Optional[str] = None,
+        timezone: Optional[str] = None,
+    ) -> Optional[Dict]:
```
Equidem, line 35: We should make a decision about the optionality of the response for all the endpoint clients, so that it is consistent.

src/apify_client/clients/resource_clients/schedule.py
```diff
@@ -0,0 +1,83 @@
...
+    def get_log(self) -> Optional[List]:
+        """Return log for the given schedule.
+
+        https://docs.apify.com/api/v2#/reference/schedules/schedule-log/get-schedule-log
+
+        Returns:
+            Retrieved log of the given schedule
+        """
+        try:
+            response = self.http_client.call(
```
Equidem, line 74: As mentioned in the task clients, I think we should standardize this and make a method for it in ResourceClient.

src/apify_client/clients/resource_clients/schedule_collection.py
```diff
@@ -0,0 +1,63 @@
...
+            is_exclusive: When set to true, don't start actor or actor task if it's still running from the previous schedule.
+            name: The name of the schedule to create.
+            actions: Actors or tasks that should be run on this schedule. See the API documentation for exact structure.
+            description: Description of this schedule
+            timezone: Timezone in which your cron expression runs (TZ database name from https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)
+
+        Returns:
+            The created schedule.
+        """
+        return self._create({
```
Equidem, line 55: Why do you not use the iterative approach using _snake_case_to_camel_case for these cases?

src/apify_client/clients/resource_clients/schedule.py
mhamas: I agree. And actually, `self._update` returns `Dict`, so I can use `Dict` here

src/apify_client/clients/resource_clients/schedule_collection.py
mhamas: Good catch, don't know why I missed this :D. Thx

src/apify_client/clients/resource_clients/schedule.py
mhamas: Let's discuss and if need be create a separate PR for that (refactoring should be pretty easy for that).

Equidem (approved):

fnesveda (approved):

PR #30, opened 2021-02-10
Feature/webhook client

Equidem:
Clients for manipulating webhooks and their dispatches.

### DIFF
--- a/src/apify_client/_utils.py
+++ b/src/apify_client/_utils.py
@@ -133,6 +133,7 @@ def _catch_not_found_or_throw(exc: ApifyApiError) -> None:
 
 def _snake_case_to_camel_case(str_snake_case: str) -> str:
     """Convert string in snake case to camel case.
+
     >>> _snake_case_to_camel_case("")
     ''
     >>> _snake_case_to_camel_case("making")
--- a/src/apify_client/_utils.py
+++ b/src/apify_client/_utils.py
@@ -4,7 +4,7 @@ import re
 import time
 from datetime import datetime, timezone
 from http import HTTPStatus
-from typing import Any, Callable, Dict, TypeVar, cast
+from typing import Any, Callable, Dict, List, TypeVar, cast
 
 from ._errors import ApifyApiError
 
@@ -53,6 +53,13 @@ def _parse_date_fields_internal(data: object, max_depth: int = PARSE_DATE_FIELDS
 def _pluck_data(parsed_response: Any) -> Dict:
     if isinstance(parsed_response, dict) and 'data' in parsed_response:
         return cast(Dict, parsed_response['data'])
+
+    raise ValueError('The "data" property is missing in the response.')
+
+
+def _pluck_data_as_list(parsed_response: Any) -> List:
+    if isinstance(parsed_response, dict) and 'data' in parsed_response:
+        return cast(List, parsed_response['data'])
 
     raise ValueError('The "data" property is missing in the response.')
 
@@ -122,3 +129,20 @@ def _catch_not_found_or_throw(exc: ApifyApiError) -> None:
         raise exc
 
     return None
+
+
+def _snake_case_to_camel_case(str_snake_case: str) -> str:
+    """Convert string in snake case to camel case.
+    >>> _snake_case_to_camel_case("")
+    ''
+    >>> _snake_case_to_camel_case("making")
+    'making'
+    >>> _snake_case_to_camel_case("making_the_web_programmable")
+    'makingTheWebProgrammable'
+    >>> _snake_case_to_camel_case("making_the_WEB_programmable")
+    'makingTheWebProgrammable'
+    """
+    return "".join([
+        part.capitalize() if i > 0 else part
+        for i, part in enumerate(str_snake_case.split("_"))
+    ])
--- a/src/apify_client/clients/base/base_client.py
+++ b/src/apify_client/clients/base/base_client.py
@@ -47,3 +47,15 @@ class BaseClient:
         merged_params = self.params.copy()
         merged_params.update(kwargs)
         return merged_params
+
+    def _init_options(self, **kwargs: Any) -> Dict:
+        options = {
+            "base_url": self.base_url,
+            "http_client": self.http_client,
+            "resource_id": self.resource_id,
+            "resource_path": self.resource_path,
+            "params": self.params,
+        }
+
+        options.update(kwargs)
+        return options
--- a/src/apify_client/clients/base/base_client.py
+++ b/src/apify_client/clients/base/base_client.py
@@ -52,8 +52,6 @@ class BaseClient:
         options = {
             "base_url": self.url,
             "http_client": self.http_client,
-            "resource_id": self.resource_id,
-            "resource_path": self.resource_path,
             "params": self.params,
         }
 
--- a/src/apify_client/clients/base/base_client.py
+++ b/src/apify_client/clients/base/base_client.py
@@ -48,9 +48,9 @@ class BaseClient:
         merged_params.update(kwargs)
         return merged_params
 
-    def _init_options(self, **kwargs: Any) -> Dict:
+    def _sub_resource_init_options(self, **kwargs: Any) -> Dict:
         options = {
-            "base_url": self.base_url,
+            "base_url": self.url,
             "http_client": self.http_client,
             "resource_id": self.resource_id,
             "resource_path": self.resource_path,
--- a/src/apify_client/clients/base/base_client.py
+++ b/src/apify_client/clients/base/base_client.py
@@ -46,7 +46,7 @@ class BaseClient:
     def _params(self, **kwargs: Any) -> Dict:
         return {
             **self.params,
-            **kwargs
+            **kwargs,
         }
 
     def _sub_resource_init_options(self, **kwargs: Any) -> Dict:
@@ -58,5 +58,5 @@ class BaseClient:
 
         return {
             **options,
-            **kwargs
+            **kwargs,
         }
--- a/src/apify_client/clients/base/base_client.py
+++ b/src/apify_client/clients/base/base_client.py
@@ -44,9 +44,10 @@ class BaseClient:
         return self.url
 
     def _params(self, **kwargs: Any) -> Dict:
-        merged_params = self.params.copy()
-        merged_params.update(kwargs)
-        return merged_params
+        return {
+            **self.params,
+            **kwargs
+        }
 
     def _sub_resource_init_options(self, **kwargs: Any) -> Dict:
         options = {
@@ -55,5 +56,7 @@ class BaseClient:
             "params": self.params,
         }
 
-        options.update(kwargs)
-        return options
+        return {
+            **options,
+            **kwargs
+        }
--- a/src/apify_client/clients/resource_clients/webhook.py
+++ b/src/apify_client/clients/resource_clients/webhook.py
@@ -1,10 +1,12 @@
 from typing import Any, Dict, List, Optional
 
+from ..._utils import _snake_case_to_camel_case
 from ..base.resource_client import ResourceClient
 from .webhook_dispatch_collection import WebhookDispatchCollectionClient
 
 
-def prepare_webhook_representation(
+def _prepare_webhook_representation(
+    *,
     event_types: Optional[List] = None,
     request_url: Optional[str] = None,
     payload_template: Optional[str] = None,
@@ -17,18 +19,13 @@ def prepare_webhook_representation(
     is_ad_hoc: Optional[bool] = None,
 ) -> Dict:
     """Prepare webhook dictionary representation for clients."""
-    webhook: Dict[str, Any] = {}
-
-    if event_types is not None:
-        webhook['eventTypes'] = event_types
-
-    if request_url is not None:
-        webhook['requestUrl'] = request_url
-
-    if payload_template is not None:
-        webhook['payloadTemplate'] = payload_template
+    webhook: Dict[str, Any] = {
+        _snake_case_to_camel_case(key): value
+        for key, value in locals().items() if value is not None and key not in ["actor_run_id", "actor_task_id", "actor_id"]
+    }
 
     condition = {}
+
     if actor_run_id is not None:
         condition['actorRunId'] = actor_run_id
         webhook['isAdHoc'] = True
@@ -38,19 +35,7 @@ def prepare_webhook_representation(
         condition['actorId'] = actor_id
 
     if condition != {}:
-        webhook['condition'] = condition
-
-    if ignore_ssl_errors is not None:
-        webhook['ignoreSslErrors'] = ignore_ssl_errors
-
-    if do_not_retry is not None:
-        webhook['doNotRetry'] = do_not_retry
-
-    if idempotency_key is not None:
-        webhook['idempotencyKey'] = idempotency_key
-
-    if is_ad_hoc is not None:
-        webhook['isAdHoc'] = is_ad_hoc
+        webhook["condition"] = condition
 
     return webhook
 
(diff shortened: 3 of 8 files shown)


src/apify_client/clients/base/base_client.py
```diff
@@ -47,3 +47,13 @@ def _params(self, **kwargs: Any) -> Dict:
...
+
+    def _sub_resource_init_options(self, **kwargs: Any) -> Dict:
+        options = {
+            "base_url": self.url,
+            "http_client": self.http_client,
+            "params": self.params,
+        }
+
+        options.update(kwargs)
+        return options
```
mhamas, line 59: Wouldn't the following be easier to read?
```
return {
     ...options,
     ...kwargs
}
```

src/apify_client/clients/resource_clients/webhook_dispatch.py
```diff
@@ -0,0 +1,21 @@
...
+    """Sub-client for querying information about a webhook dispatch."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the WebhookDispatchClient."""
+        super().__init__(*args, resource_path='webhook-dispatches', **kwargs)
+
+    def get(self) -> Optional[Dict]:
+        """Retrieve the webhook dispatch.
+
+        https://docs.apify.com/api/v2#/reference/webhook-dispatches
```
mhamas, line 16: Wrong URL: `https://docs.apify.com/api/v2#/reference/webhook-dispatches/webhook-dispatch-object/get-webhook-dispatch`

src/apify_client/clients/resource_clients/webhook_dispatch_collection.py
```diff
@@ -0,0 +1,26 @@
...
+
+from ..base.resource_collection_client import ResourceCollectionClient
+
+
+class WebhookDispatchCollectionClient(ResourceCollectionClient):
+    """Sub-client for listing webhook dispatches."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the WebhookDispatchCollectionClient."""
+        resource_path = kwargs.pop('resource_path', 'webhook-dispatches')
```
mhamas, line 11: Why is the `pop` needed here? Can't you simply hard code `webhook-dispatches` on the next line in `super().__init__(...)`? That's how it's done in e.g. dataset collection client.

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,130 @@
...
+        event_types: Optional[List] = None,
+        request_url: Optional[str] = None,
+        payload_template: Optional[str] = None,
+        actor_id: Optional[str] = None,
+        actor_task_id: Optional[str] = None,
+        actor_run_id: Optional[str] = None,
+        ignore_ssl_errors: Optional[bool] = None,
+        do_not_retry: Optional[bool] = None,
+        is_ad_hoc: Optional[bool] = None,
+    ) -> Dict:
```
mhamas, line 86: isn't this `Optional[Dict?`

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,130 @@
...
+                                           failure.
+            is_ad_hoc (bool, optional): Set to True if you want the webhook to be triggered only the first time the
+                                        condition is fulfilled. Only applicable when actor_run_id is filled.
+
+        Returns:
+            The updated webhook
+        """
+        parameters = locals()
+        parameters.pop('self')
+        webhook = prepare_webhook_representation(**parameters)
```
mhamas, line 110: I'm quite fond of dict comprehension (find it easy to read as it's very explicit), what about the following
```
webhook = prepare_webhook_representation(**{
    key: value
    for key, value of locals().items() if key != 'self'
})
```

mhamas: Btw, no hard opinion here, your way is also fine.

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,130 @@
...
+        """Delete the webhook.
+
+        https://docs.apify.com/api/v2#/reference/webhooks/webhook-object/delete-webhook
+        """
+        return self._delete()
+
+    def dispatches(self) -> WebhookDispatchCollectionClient:
+        """Get dispatches of the webhook.
+
+        https://docs.apify.com/api/v2#/reference/webhooks/dispatches-collection
```
mhamas, line 123: Wrong URL: `https://docs.apify.com/api/v2#/reference/webhooks/dispatches-collection/get-collection`

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,130 @@
+from typing import Any, Dict, List, Optional
+
+from ..base.resource_client import ResourceClient
+from .webhook_dispatch_collection import WebhookDispatchCollectionClient
+
+
+def prepare_webhook_representation(
+    event_types: Optional[List] = None,
```
mhamas, line 8: lets have `*` args as the first arg in this list to force user to user named params.

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,130 @@
+from typing import Any, Dict, List, Optional
+
+from ..base.resource_client import ResourceClient
+from .webhook_dispatch_collection import WebhookDispatchCollectionClient
+
+
+def prepare_webhook_representation(
```
mhamas, line 7: I'd rename to `_prepare_webhook_representation` since this is just internal function.

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,130 @@
...
+    if do_not_retry is not None:
+        webhook['doNotRetry'] = do_not_retry
+
+    if idempotency_key is not None:
+        webhook['idempotencyKey'] = idempotency_key
+
+    if is_ad_hoc is not None:
+        webhook['isAdHoc'] = is_ad_hoc
+
+    return webhook
```
mhamas, line 55: Look at how I did something similar in https://github.com/apify/apify-client-python/pull/29/files. Basically, I introduced function that converts snake case to camel case. And then, when preparing these args, you can do something like

```
const new_dict = {}
for key, value of locals().entries():
    ...some custom logic you might have...
    new_dict[snake_case_to_camel_case(key)] = value;
```

It's much less verbose and easier to read imho.

Also, in you case, I'd handle `condition` independently of this loop, construct it explicitly and then just assign to the new dict.

src/apify_client/clients/resource_clients/webhook_collection.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def list(self, *, limit: Optional[int] = None, offset: Optional[int] = None, desc: Optional[bool] = None) -> Dict:
+        """List the available webhooks.
+
+        https://docs.apify.com/api/v2#/reference/webhooks/webhook-collection/get-list-of-webhooks
+
+        Args:
+            limit (int, optional): How many webhooks to retrieve
+            offset (int, optional): What webhook to include as first when retrieving the list
+            desc (bool, optional): Whether to sort the webhooks in descending order based on their date of creation
```
mhamas, line 22: btw, these types in args are cool. I wasn't doing it before but I will in the future. Thx

src/apify_client/clients/resource_clients/webhook_collection.py
```diff
@@ -0,0 +1,70 @@
...
+
+        Returns:
+            The list of available webhooks matching the specified filters.
+        """
+        return self._list(limit=limit, offset=offset, desc=desc)
+
+    def create(
+        self,
+        event_types: List,
```
mhamas, line 31: I would put `*` arg after `self`

src/apify_client/clients/resource_clients/webhook_collection.py
```diff
@@ -0,0 +1,70 @@
...
+        actor_task_id: Optional[str] = None,
+        actor_run_id: Optional[str] = None,
+        ignore_ssl_errors: Optional[bool] = None,
+        do_not_retry: Optional[bool] = None,
+        idempotency_key: Optional[str] = None,
+        is_ad_hoc: Optional[bool] = None,
+    ) -> Dict:
+        """Create a new webhook.
+
+        You have to specify exactly one out of actor_id, actor_task_id or actor_run_id.
```
mhamas, line 44: If they don't API errors out?

mhamas (changes requested):
In general looks fine. Please see some smaller comments.

src/apify_client/clients/resource_clients/webhook_dispatch_collection.py
Equidem: You can't, because you have to get resource_path out from the dictionary if it is there, otherwise it will lead to an error because you would be passing one named argument twice.

src/apify_client/clients/resource_clients/webhook.py
Equidem: Why?
This is done the same way e.g. in Dataset.

src/apify_client/clients/resource_clients/webhook.py
Equidem: Your way requires the program to go through the dictionary once more than mine, but that's basically irrelevant since it is in no way a performance bottleneck, of course, just a note for when it would be important.
I also find my way easier to read.

src/apify_client/clients/resource_clients/webhook.py
Equidem: Good point, I am not used to the underscore-before-function-name notation.

src/apify_client/clients/resource_clients/webhook.py
Equidem: Good point.

src/apify_client/clients/resource_clients/webhook_collection.py
Equidem: Haven't tested it, but am 99% sure that if you don't include any, it raises an error, whereas more than one will get saved no problem. However,  specifying more than one has no function and might lead to unexpected behavior.

src/apify_client/clients/resource_clients/webhook_dispatch.py
Equidem: Will fix, thanks.

src/apify_client/clients/resource_clients/webhook.py
Equidem: Will fix, thanks.

src/apify_client/clients/resource_clients/webhook.py
Equidem: Users shouldn't use this function at all.

src/apify_client/clients/resource_clients/webhook_collection.py
Equidem: Right, forgot to do that, thanks.

src/apify_client/clients/base/base_client.py
Equidem: Just as the other similar note - slightly less efficient, though it doesn't matter here. 
I will change this one, since your way is prettier in this case :)

src/apify_client/clients/resource_clients/webhook.py
mhamas: I just thought it can actually return `None` if something goes wrong. But maybe not

src/apify_client/clients/resource_clients/webhook.py
mhamas: yes, but it will force us to use named args, which is good practice :-)

src/apify_client/clients/resource_clients/webhook.py
mhamas: me neither, but seems we are using it so let's try to be consistent

src/apify_client/clients/resource_clients/webhook.py
Equidem: I would think that it would raise an error if something goes wrong.

src/apify_client/_utils.py
```diff
@@ -57,6 +57,13 @@ def _pluck_data(parsed_response: Any) -> Dict:
     raise ValueError('The "data" property is missing in the response.')
 
 
+def _pluck_data_as_list(parsed_response: Any) -> List:
+    if isinstance(parsed_response, dict) and 'data' in parsed_response:
+        return cast(List, parsed_response['data'])
+
+    raise ValueError('The "data" property is missing in the response.')
+
+
```
mhamas, line 66: are you using this?

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,116 @@
...
+
+    if actor_run_id is not None:
+        condition['actorRunId'] = actor_run_id
+        webhook['isAdHoc'] = True
+    elif actor_task_id is not None:
+        condition['actorTaskId'] = actor_task_id
+    elif actor_id is not None:
+        condition['actorId'] = actor_id
+
+    if condition != {}:
```
mhamas, line 37: I think `if condition:` should be enough

src/apify_client/clients/resource_clients/webhook.py
```diff
@@ -0,0 +1,116 @@
...
+        do_not_retry: Optional[bool] = None,
+        is_ad_hoc: Optional[bool] = None,
+    ) -> Dict:
+        """Update the webhook.
+
+        https://docs.apify.com/api/v2#/reference/webhooks/webhook-object/update-webhook
+
+        Args:
+            event_types (list, optional): List of event types that should trigger the webhook.
+                                Present in the client constants as WebhookEventType. At least one is required.
```
mhamas, line 79: Couple of lines under this, you format overflowing text such as the description is aligned. Let's do it that way so we are consistent.

src/apify_client/clients/resource_clients/webhook_dispatch_collection.py
mhamas: can you please leave a comment here? Thanks

mhamas (changes requested):
Looking good, just a couple of more things I noticed after the update.

src/apify_client/_utils.py
Equidem: Shouldn't things like this be done where the request is made? (ResourceClient, I think)
Seems it could be done universally for all requests.

src/apify_client/clients/resource_clients/webhook.py
Equidem: That's against one of the principles of clean code. If you know what precisely you are testing for, you should always test precisely for that and not use a more general condition, since that way you just invite unexpected behavior.

src/apify_client/_utils.py
Equidem: @fnesveda

Equidem: Or did I misunderstand you and your question is why the function is in _utils.py? If that's the case, then I am not completely sure, but it shouldn't matter.

Equidem: Oh, I remember now. To decrease the probability of merging conflicts, I stole the entire file from one of your branches.

fnesveda (approved):
Looking good! I'm glad that I went second with the review after Maťo, I had a lot easier job :D

src/apify_client/_utils.py
mhamas: The question was why this function is here because it seems not be used in this branch. Now I understand you grabbed a file from different branch. Still, please remove this function if you don't use it in this code. You have no guarantee this file in that other branch changes and they you'd accidentally check-in dead code. Thx

src/apify_client/clients/resource_clients/webhook.py
mhamas: Ok, good argument, thx

mhamas (approved):
Looks good, please remove one function with dead code (`_pluck_data_as_list`). Apart from that, good to go.

fnesveda:
@mhamas @Equidem  I merged master into this branch and the extra function got merged properly, so no need to do anything.

PR #31, opened 2021-02-12
Task collection subclient

mhamas:

PR #32, opened 2021-02-12
Sub-client for single Task

mhamas:
Note that the subclient can't be properly finished until Run subclient is done. Hence some methods still throw errors. As far as I'm concerned, we can merge it with unfinished methods and we will implement them later.


src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+        """Delete the task.
+
+        https://docs.apify.com/api/v2#/reference/actor-tasks/task-object/delete-task
+        """
+        return self._delete()
+
+    def start(
+        self,
+        *,
+        task_input: Optional[Dict[str, Any]] = None,
```
mhamas, line 73: I'm assuming everywhere that input is `Dict` only. @fnesveda is that correct? Thx

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+        build: Optional[str] = None,
+        memory_mb: Optional[int] = None,
+        timeout_secs: Optional[int] = None,
+    ) -> Dict:
+        """Update the task with specified fields.
+
+        https://docs.apify.com/api/v2#/reference/actor-tasks/task-object/update-task
+
+        Args:
+            name (string, optional): Name of the task
```
Equidem, line 41: Should be (str, optional)

src/apify_client/_utils.py
```diff
@@ -122,3 +125,51 @@ def _catch_not_found_or_throw(exc: ApifyApiError) -> None:
...
+
+
+def _filter_out_none_values_recursively(dictionary: Dict) -> Dict:
+    """Return copy of the dictionary, recursivelhy omitting all keys for which values are None.
+
+    >>> _filter_out_none_values_recursively({"k1": "v1", "k2": None, "k3": {"k4": "v4", "k5": None}})
+    {'k1': 'v1', 'k3': {'k4': 'v4'}}
+    """
+    return {
+        k: v if not isinstance(v, Dict) else _filter_out_none_values_recursively(v)
```
Equidem, line 172: Recursion is usually pretty unsafe to use (due to the possibility of exceeding max call stack depth) and incurs the unnecessary overhead of a function call. It's not really a problem here due to the small scale of the intended dictionaries, of course. It is more a matter of principle.

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+            build (str, optional): Specifies the actor build to run. It can be either a build tag or build number.
```
Equidem, line 42: All the comments around are worded as a definition (e.g. "Name of the task"), while this one seems more like an "explanation" (begins with a verb).

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+                                   By default, the run uses the build specified in the task settings (typically latest).
+            memory_mb (int, optional): Memory limit for the run, in megabytes. By default, the run uses a memory limit specified in the task settings.
```
Equidem, line 44: Maybe we could call this one memory_mbytes? The unit would then be clear from the name without the need to read the comment.

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+        self,
+        *,
+        task_input: Optional[Dict[str, Any]] = None,
+        build: Optional[str] = None,
+        memory_mb: Optional[int] = None,
+        timeout_secs: Optional[int] = None,
+        wait_for_finish: Optional[int] = None,
+        webhooks: Optional[List[Dict]] = None,
+    ) -> Dict:
+        """Start a task and immediately return the Run object.
```
Equidem, line 80: "Start the task ..." would make it a little bit clearer.

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Args:
+            task_input (dict, optional): Task input dictionary
+            build (str, optional): Specifies the actor build to run. It can be either a build tag or build number.
+                                   By default, the run uses the build specified in the task settings (typically latest).
+            memory_mb (int, optional): Memory limit for the run, in megabytes. By default, the run uses a memory limit specified in the task settings.
+            timeout_secs: (int, optional): Optional timeout for the run, in seconds. By default, the run uses timeout specified in the task settings.
+            wait_for_finish: (bool, optional): The maximum number of seconds the server waits for the run to finish.
+                                               By default, it is 0, the maximum value is 300.
+            webhooks (list, optional): Specifies optional webhooks associated with the actor run, which can be used to receive a notification
```
Equidem, line 92: The same issue as with the "build" parameter comments, also could do without the "Note:", since the entire comment is a note ;)

Equidem: Also, it wouldn't be clear what format the webhooks should be in if I were a user.

src/apify_client/_utils.py
```diff
@@ -122,3 +125,51 @@ def _catch_not_found_or_throw(exc: ApifyApiError) -> None:
         raise exc
 
     return None
+
+
+def _encode_json_to_base64(data: JSONSerializable) -> bytes:
```
Equidem, line 130: Seems to be a misnomer - the function doesn't encode a JSON string, but a JSON-serializable types of objects. Same with the _decode_base64_json function, which also lacks "to" between base64 and json.

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        request_params = self._params(
+            build=build,
+            memory=memory_mb,
+            timeout=timeout_secs,
+            waitForFinish=wait_for_finish,
+            webhooks=_encode_json_to_base64(webhooks) if webhooks is not None else [],
+        )
+
+        response = self.http_client.call(
```
Equidem, line 107: Not sure, but shouldn't this rather be "not finished yet" too? Seems everything in this function should instead be one call to the RunClient, or we will repeat the same code several times.

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+        #     memory_mb=memory_mb,
+        #     timeout_secs=timeout_secs,
+        #     wait_for_finish=wait_for_finish,
+        #     webhooks=webhooks,
+        # )
+
+        # TODO - retrieve the run using Run client and wait on it
+
+    def get_input(self) -> Optional[Dict]:
+        """Retrieve the input for this task.
```
Equidem, line 162: Maybe add default to the comment? ("Retrieve the default input for this task.")

Equidem: Same in the "update_input" function.

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+
+        https://docs.apify.com/api/v2#/reference/actor-tasks/task-input-object/get-task-input
+
+        Returns:
+            Retrieved task input
+        """
+        try:
+            response = self.http_client.call(
```
Equidem, line 170: Maybe there could be a BaseClient or ResourceClient function for this call? (if this kind occurs in other endpoints too)

Equidem: Same for the update_input function, of course.

src/apify_client/clients/resource_clients/task.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        response = self.http_client.call(
+            url=self._url('input'),
+            method='PUT',
+            params=self._params(),
+            json=task_input,
+        )
+        return cast(Dict, response.json())
+
+    def last_run(self, *, status: Optional[ActorJobStatus] = None) -> Any:
```
Equidem, line 196: Maybe there could be an option to differentiate between the run that was started last and the run that finished last?

Equidem (changes requested):
Good work, my comments are mainly just nitpicks about naming and comments.

src/apify_client/_utils.py
mhamas: not sure what you mean by this, can you please elaborate, ideally with an example_

src/apify_client/_utils.py
mhamas: Good point, but I don't really think it should be a problem here. Do you have strong opinion on this? I guess we can rewrite it using queue / stack, but I don't think it's really needed.

src/apify_client/clients/resource_clients/task.py
mhamas: Same comment as https://github.com/apify/apify-client-python/pull/31#discussion_r575993702

src/apify_client/clients/resource_clients/task.py
mhamas: JS client does it explicitly (https://github.com/apify/apify-client-js/blob/master/src/resource_clients/task.js#L78-L94), I've been mimicking that.

src/apify_client/clients/resource_clients/task.py
mhamas: Let's leave this to a separate PR if we decided to do that

src/apify_client/clients/resource_clients/task.py
mhamas: Is there this functionality in JS client (@mnmkng )? I am not aware of it.

mnmkng: The `lastRun` uses the [last run API endpoint](https://docs.apify.com/api/v2#/reference/actor-tasks/last-run-object-and-its-storages). There's no client magic behind it and I wouldn't add any at this point.

src/apify_client/clients/resource_clients/task.py
mhamas: Renamed to `mbytes`

src/apify_client/_utils.py
Equidem: First of all, I think the _decode_base64_json should be _decode_base64_to_json.
However, my main complain is that _encode_json_to_base64 doesn't actually take JSON as an input, it usually takes a dictionary, doesn't it?

mhamas: Renamed...as for `_encode_json_to_base64`, it can take any `JSONSerializable` value, look at the doc tests in `_decode_base64_to_json` where this is featured

src/apify_client/clients/resource_clients/task.py
Equidem: Sure, but that doesn't mean it can't be improved on.

```
         response = self.http_client.call(
            url=self._url('runs'),
            method='POST',
            headers={'content-type': 'application/json; charset=utf-8'},
            json=task_input,
            params=request_params,
        )

        return _parse_date_fields(_pluck_data(response.json()))
```
This could be a single short function call instead of one long one and one short one. And that's not even mentioning that if we ever decide to change the way we parse the response for instance, we would have to go through all these files.

src/apify_client/clients/resource_clients/task.py
Equidem: Okey. It is just a little confusing what run is actually the last, so maybe just mention it in the function description?

src/apify_client/clients/resource_clients/task.py
mhamas: It would be easy to create such a function, but it would need to accept many arguments (at least `url`, `method`, `json`, `params`, `data`) and would execute 2 lines of code. And also, what would we call that function? `_call_get_data_and_parse_dates`?  I can't think of a good short name :D.

I'm not convinced this would help us, it feels more like passing around arguments. But I'm open for an another opinion,  @fnesveda, @mnmkng ? Thx

src/apify_client/clients/resource_clients/task.py
mhamas: Good idea, I added it to the function description. Thx

src/apify_client/_utils.py
Equidem: I think we can keep the recursion here, since it should never become a problem. Just be careful where you use the function.

src/apify_client/clients/resource_clients/task.py
Equidem: _execute_request(...) ?

fnesveda (approved):

PR #38, opened 2021-04-21
Improve list pagination with new ListPage class

fnesveda:
Currently, the `list()` methods on the various resource collection clients return a dictionary with the keys `items`, `offset`, `limit`, `count` and `total`. This is pretty hard to work with, it's hard to document and you don't get any IDE suggestions for the dictionary members etc.
I replaced that implementation with returning instances of a new `ListPage` class, which has the same members as the original dictionary, but is well documented and typed, so it's easier to work with.

The formatting of the documentation is becoming pretty complex, perhaps we should have used a completely custom documentation renderer from the start. Too late, it would be a lot of work to switch and since it works, there's no good reason to replace it right now, but perhaps in the future, if we ever change the documentation format, we should start from scratch and use a custom renderer.


src/apify_client/_utils.py
```diff
@@ -201,3 +201,26 @@ def _encode_key_value_store_record_value(value: Any, content_type: Optional[str]
         value = json.dumps(value, ensure_ascii=False, indent=2).encode("utf-8")
 
     return (value, content_type)
+
+
+class ListPage:
+    """A single page of items returned from a list() method."""
+
+    #: list: List of returned objects on this page
```
mhamas, line 209: What is this notation?

src/apify_client/_utils.py
```diff
@@ -201,3 +201,26 @@ def _encode_key_value_store_record_value(value: Any, content_type: Optional[str]
...
+    #: int: Total number of objects matching the API call criteria
+    total: int
+
+    def __init__(self, data: Dict) -> None:
+        """Initialize a ListPage instance from the API response data."""
+        self.items = data['items'] if 'items' in data else []
+        self.offset = data['offset'] if 'offset' in data else 0
+        self.limit = data['limit'] if 'limit' in data else 0
+        self.count = data['count'] if 'count' in data else len(self.items)
+        self.total = data['total'] if 'total' in data else self.offset + self.count
```
mhamas, line 226: I think it'd be more canonical to have proper `__init__` accepting individual arguments and use `@classmethod` to create `ListPage` from dict.

```
@classmethod
def from_dict(cls, data):
    return cls(...individual values from data...)
    
To simplify declaring members and then explicitly writing init, you can also use `attrs` package (https://www.attrs.org/en/stable/) which helps you to define this with minimum boilerplate.

mhamas (approved):
In general looks good to me, please see comments.

PR #39, opened 2021-04-21
Send API token in the Authorization header instead of as a request parameter

fnesveda:
The API server now supports authorizing with a token sent in the `Authorization` header for improved security, so this PR adds support for sending it that way.


mhamas (approved):

drobnikj (approved):
Cool, let's add me to more reviews about python client, I would like to learn about it more.

PR #40, opened 2021-05-13
Set up publishing to PyPI via GitHub Actions

fnesveda:
I've set up publishing of this package to PyPI via GitHub Actions. It works similarly as the publishing of `apify-client-js` to NPM:
- the package version is single-sourced from `src/apify_client/_version.py`
- on a commit to `master`, a new beta release is created and published to PyPI, automatically appending a beta number to the version, incrementing the beta number from the last published beta release for the current version
- when a new GitHub Release is created, a new stable version is published to PyPI
- if a version same as in `src/apify_client/_version.py` was already published, the publication fails
- if `CHANGELOG.md` was not updated for the published version, the publication fails

It's pretty hard to test this properly live without polluting the PyPI listing for this package, but I've tested and debugged this exact same release process on TestPyPI with a dummy package in a private repo, and everything worked fine, so I'm reasonably confident this will work too.


mnmkng (approved):
The workflow looks good to me 👍

mhamas (approved):

PR #45, opened 2021-06-09
Update development dependencies, add types-request dev dependency

fnesveda:
I updated all the packages needed for development to their latest versions. `mypy >= 0.900` and does not include the types for popular libraries anymore, so we have to include them as a separate package, so I also added the `types-request` package to fix the arising issues.


src/apify_client/_version.py
```diff
@@ -1 +1 @@
-__version__ = '0.0.1'
+__version__ = '0.1.0'
```
mhamas, line 1: Climbing the version ladder!

mhamas (approved):

drobnikj (approved):

Equidem (approved):

PR #46, opened 2021-06-09
Enforce use of single-quotes via lint plugin

fnesveda:
I added the package `flake8-quotes` as a development dependency, it enforces the use of single-quotes by default through linting, and I fixed all the violations that were present in the code, as it makes it a bit cleaner and unified.


Equidem (approved):

drobnikj (approved):

mhamas (approved):
great, thx

PR #47, opened 2021-06-09
Enforce use of Enum items in method arguments

fnesveda:
In some methods in the resource clients, we had the type of some arguments set to some `Enum`, but we actually worked with the arguments as if they were strings. This PR corrects it so that we're actually working with those arguments as `Enum` items, and updates the documentation of those arguments to point to those `Enum`s.

The PR also adds the description of those `Enum`s to the documentation, which required a small rework of how the formatting of the documentation works, adding to the mess that that is.

I've also renamed the `_consts` submodule to just `consts`, so that users can import those `Enum`s from those consts like `from apify_client.consts import ActorJobStatus` and use the `Enum` items in the client methods.


docs/res/format_docs.py
```diff
@@ -25,6 +25,13 @@
     (r'    ', r'  '),  # indent with 2 spaces instead of 4
 ]
 
+# Add fragment link to parameters of our custom types (I know this would be better generic, but it would be really hard to write)
+custom_types = ['ActorJobStatus', 'ActorSourceType', 'WebhookEventType']
```
mhamas, line 29: maybe capital `CUSTOM_TYPES`?

mhamas (approved):
Thanks!

docs/res/format_docs.py
fnesveda: Makes sense, updated :)

Equidem (approved):

drobnikj (approved):

PR #48, opened 2021-06-12
Docs: fix link

dragonraid:


fnesveda (approved):

fnesveda:
Good catch, thanks! I'll rebase this and point this into the latest feature branch so that the lint tests are working, they're broken because of a library update not properly following semver, and there is already a fix for it in a PR.

PR #80, opened 2022-11-07
Unified methods for streaming resources

fnesveda:
With the implementation of the Apify SDK for Python, we will have to do some bigger changes to the Python Client, so now is a good time to take some time and get rid of some old problems, before we are stuck with them forever. One of them is that streaming methods for some resources (dataset items, key-value store records and log contents) are a mess.

This unifies their behavior, and makes them safer by forcing them to return a context-managed object, so the streaming response is automatically closed after use.

It also unifies the naming and behavior of methods for fetching the resource contents as raw bytes, by naming them all in the `get_as_bytes` style, instead of the varying names and usages they had before.

### DIFF
--- a/src/apify_client/clients/resource_clients/key_value_store.py
+++ b/src/apify_client/clients/resource_clients/key_value_store.py
@@ -1,4 +1,6 @@
-from typing import Any, Dict, Optional
+import warnings
+from contextlib import contextmanager
+from typing import Any, Dict, Iterator, Optional
 
 from ..._errors import ApifyApiError
 from ..._utils import (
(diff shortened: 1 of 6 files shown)


src/apify_client/clients/resource_clients/key_value_store.py
```diff
@@ -115,6 +140,71 @@ def get_record(self, key: str, *, as_bytes: bool = False, as_file: bool = False)
...
+
+            yield {
+                'key': key,
+                'value': response,
+                'content_type': response.headers['content-type'],
+            }
+
+        except ApifyApiError as exc:
+            _catch_not_found_or_throw(exc)
+            yield None
```
mhamas, line 203: You don't have this catch for dataset streaming, is that intentional?

mhamas (approved):
looks good to me, one comment to consider.

src/apify_client/clients/resource_clients/key_value_store.py
fnesveda: Yup, that's intentional. This logic (catching in k-v store, not catching in dataset) is the same in the JS client, so when originally implementing the Python client, I just copied it over to keep it consistent.

The logic is not great:
- fetching a non-existent dataset returns None
- fetching a non-existent key-value store returns None
- fetching a non-existent key from an existent key-value store returns None
- fetching a non-existent key from a non-existent key-value store returns None
- fetching items from a non-existent dataset throws an error

The last two of those are inconsistent, but I think in a way they can make sense, if you think of key-value stores as folders with files, and datasets as tables.

Anyway, I would leave it like this, as to not make more breaking changes and logical differences from the JS client.

HonzaTuron (approved):

PR #82, opened 2022-11-11
Added support for asynchronous usage via `ApifyClientAsync`

fnesveda:
The Python SDK will use async methods by default, and for it to be effective, we need to support async usage in the Python Client as well.

This PR adds `ApifyClientAsync`, a sibling to `ApifyClient` with all methods being `async`. The usage and structure is the same as in `ApifyClient`, but unfortunately, a lot of the code had to be mostly duplicated, because we had to add an `await` here and there, and we had to adjust the typings a bit. At least the docs don't have to be duplicated, I made a helper decorator for that.

The downside is that now, if we add features to the client, we will have to add it to both the sync and the async version. Hopefully in the future we will remove the sync client and have only the async one.

You don't need to review every changed line, it's enough to check the changes in the main logic, not in the actual resource clients.


jirimoravcik (approved):

HonzaTuron (approved):

jirimoravcik:
BTW, have you looked at https://pypi.org/project/asgiref/#function-wrappers ? I wonder if we could just have async implementation and use "async to sync" for the sync client?

src/apify_client/_http_client.py
```diff
@@ -121,36 +169,64 @@ def _make_request(stop_retrying: Callable, attempt: int) -> httpx.Response:
...
+
+            except Exception as e:
+                if not _is_retryable_error(e):
+                    stop_retrying()
+                raise e
+
+            if response.status_code < 500 and response.status_code != HTTPStatus.TOO_MANY_REQUESTS:
+                stop_retrying()
+            raise ApifyApiError(response, attempt)
+
```
mhamas, line 225: can you comment on this logic when checking status codes pls? (`<300`, `<500`).

mhamas (approved):
looks good to me

PR #85, opened 2022-11-22
Catch up to JS client

jirimoravcik:
The only thing omitted is the `title` parameter for dataset, kvs and rq, I only added it for schedule and task. This is because it would be inconsistent in the current state i.e. you cannot create a storage with a `title` but you can do it via update. See https://github.com/apify/apify-client-js/pull/271

Should be merged after the logger PR #86 since I merged it into this branch.


fnesveda (changes requested):
Nice, looks great! I just had a couple of small comments.

fnesveda (changes requested):
Nice! Oh and one more thing, can you add the changes to changelog, please?

mhamas (approved):
looks good to me, thanks

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ Changelog
 - added method to update status message for a run
 - added option to set up webhooks for actor builds
 - added logger with basic debugging info
+- added support for `schema` parameter in `getOrCreate` method for datasets and key-value stores
```
fnesveda, line 21: We use `snake_case` here 🙂 
```suggestion
- added support for `schema` parameter in `get_or_create` method for datasets and key-value stores
```

fnesveda (changes requested):
Almost perfect, last nitpick now 😄

jirimoravcik:
> Almost perfect, last nitpick now 😄

yes, i'm struggling with snake case a lot atm :D

fnesveda (approved):

PR #86, opened 2022-11-29
Added logger with basic debugging info

fnesveda:
Logging was missing for a long time from the client, and it made it harder to debug the client.

This adds some basic logging to the HTTP calls, using the standard Python logger, enriched by contextual information from the client methods which are doing the HTTP calls by some funky metaclass programming.

There is a `_DebugLogFormatter` class which is great for testing, used like this:

```python
import logging
from apify_client._logging import _DebugLogFormatter

client_logger = logging.getLogger('apify_client')
client_logger.setLevel(logging.DEBUG)
handler = logging.StreamHandler()
handler.setFormatter(_DebugLogFormatter())
client_logger.addHandler(handler)
```

### DIFF
--- a/docs/res/intro.md
+++ b/docs/res/intro.md
@@ -185,3 +185,22 @@ async def main():
 
 asyncio.run(main())
 ```
+
+### Logging
+
+The library logs some useful debug information to the `apify_client` logger
+when sending requests to the Apify API.
+To have them printed out to the standard output, you need to add a handler to the logger:
+
+```python
+import logging
+apify_client_logger = logging.getLogger('apify_client')
+apify_client_logger.setLevel(logging.DEBUG)
+apify_client_logger.addHandler(logging.StreamHandler())
+```
+
+The log records have useful properties added with the `extra` argument,
+like `attempt`, `status_code`, `url`, `client_method` and `resource_id`.
+To print those out, you'll need to use a custom log formatter.
+To learn more about log formatters and how to use them,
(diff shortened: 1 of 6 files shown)


jirimoravcik (approved):

HonzaTuron (approved):

docs/res/intro.md
```diff
@@ -185,3 +185,22 @@ async def main():
...
+apify_client_logger = logging.getLogger('apify_client')
+apify_client_logger.setLevel(logging.DEBUG)
+apify_client_logger.addHandler(logging.StreamHandler())
+```
+
+The log records have useful properties added with the `extra` argument,
+like `attempt`, `status_code`, `url`, `client_method` and `resource_id`.
+To print those out, you'll need to use a custom log formatter.
+To learn more about log formatters and how to use them,
+please refer to the official Python [documentation on logging](https://docs.python.org/3/howto/logging.html#formatters).
```
mhamas, line 206: Maybe it would be beneficial here to show a simple example with custom log formatter? So the users don't need to go dig in the official docs.

mhamas (commented):
This is the biggest python magic I've every seen :D. I've never worked with context variables and meta classes, but after reading up on them the usage here looks good to me. A thanks for teaching me something new!

fnesveda:
@mhamas You're welcome! Python can be quite powerful, I'm glad we're now doing the Python SDK because I'm getting to dive deep into it. Currently about to send my first PR to CPython 😄

PR #94, opened 2023-01-06
fix: Change dataset iterate_items return type to AsyncIterator[Dict]

jirimoravcik:


fnesveda (changes requested):
Makes sense, could you also change the sync version on line 143?

PR #100, opened 2023-01-25
Write docs for async methods explicitly, get rid of `_make_async_docs` decorator

fnesveda:
Since the docs for async methods which were generated with the `_make_async_docs` decorator were not showing in the hints in VSCode, and were not processed in `pydoc-markdown`, it's better that we write the docstrings for async methods explicitly.

So that we don't have to copy-paste all the docstrings manually, I wrote a script, which uses the [Red Baron](https://github.com/PyCQA/redbaron) package, which synchronizes the docstrings from the sync methods to their analogous async methods, making some adjustments to make the docstrings work for the async methods.

I also wrote another script, which runs in CI, and just checks that the docstrings are synchronized, failing the check if they're not.

This was way easier than I was worried about, we should have done this right from the start.

CC @barjin I hope you haven't spent too much time on replicating the `_make_async_docs` decorator while rendering the docs, this should make your job a fair bit easier.

### DIFF
--- a/scripts/fix_async_docstrings.py
+++ b/scripts/fix_async_docstrings.py
@@ -1,0 +1,51 @@
+import re
+from pathlib import Path
+
+from redbaron import RedBaron  # type: ignore
+from utils import sync_to_async_docstring
+
+clients_path = Path(__file__).parent.resolve() / '../src/apify_client'
+for client_source_path in clients_path.glob('**/*.py'):
+    with open(client_source_path, 'r+') as source_file:
+        red = RedBaron(source_code=source_file.read())
+        async_class = red.find('ClassNode', name=re.compile('.*ClientAsync$'))
+        if not async_class:
+            continue
+        sync_class = red.find('ClassNode', name=async_class.name.replace('ClientAsync', 'Client'))
+        for async_method in async_class.find_all('DefNode'):
+            sync_method = sync_class.find('DefNode', name=async_method.name)
+            if isinstance(sync_method.value[0].value, str):
+                sync_docstring = sync_method.value[0].value
+                async_docstring = async_method.value[0].value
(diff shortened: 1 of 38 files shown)


scripts/fix_async_docstrings.py
```diff
@@ -0,0 +1,51 @@
+import re
+from pathlib import Path
+
+from redbaron import RedBaron  # type: ignore
```
jirimoravcik, line 4: Why type: ignore?

jirimoravcik (approved):
RedBaron bugs are interesting, I was wondering if we could call `make format` after the fix script instead of hacking the indentation inside the script

fnesveda:
@jirimoravcik Unfortunately, calling `make format` after the Red Baron execution doesn't work, because the formatting only works on valid code, but the Red Baron bugs produce invalid code (since Python is pretty strict about indentation) :(

scripts/fix_async_docstrings.py
fnesveda: Because `redbaron` doesn't have typings, and I couldn't find any type stubs (like e.g. `types-setuptools`), and `mypy` was complaining that it can't find any types for the package.

jirimoravcik: makes sense

jirimoravcik:
> @jirimoravcik Unfortunately, calling `make format` after the Red Baron execution doesn't work, because the formatting only works on valid code, but the Red Baron bugs produce invalid code (since Python is pretty strict about indentation) 🙁

unlucky

drobnikj (approved):
Nice!

I like the package name 
> RedBaron guaranteed you that it will only modify your code where you ask him to.

barjin (approved):
Great! While this might be achievable through some hacks on my side, it's definitely a huge simplification. Thanks!

PR #105, opened 2023-01-31
feat: landing page update

barjin:
Text from notion, code example taken from [here](https://docs.apify.com/apify-client-python#quick-start).

obrazek


mtrunkat:
@PerVillalva please check the copy

PR #109, opened 2023-02-02
docs: fixing (most) Notion points

barjin:
This PR solves most of the minor problems with the Python Client docs mentioned in the [Notion page](https://www.notion.so/apify/Python-specific-Docs-V2-feedback-e3f0208fd1b94d6cb1a3e84d1e875156).

### DIFF
--- a/website/patches/docusaurus-plugin-typedoc-api+2.5.1.patch
+++ b/website/patches/docusaurus-plugin-typedoc-api+2.5.1.patch
@@ -1,3 +1,27 @@
+diff --git a/node_modules/docusaurus-plugin-typedoc-api/lib/components/MemberSignatureTitle.js b/node_modules/docusaurus-plugin-typedoc-api/lib/components/MemberSignatureTitle.js
+index 0a36b74..4a2b7fb 100644
+--- a/node_modules/docusaurus-plugin-typedoc-api/lib/components/MemberSignatureTitle.js
++++ b/node_modules/docusaurus-plugin-typedoc-api/lib/components/MemberSignatureTitle.js
+@@ -23,7 +23,9 @@ function MemberSignatureTitle({
+   hideName,
+   sig
+ }) {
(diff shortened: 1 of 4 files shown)


website/patches/docusaurus-plugin-typedoc-api+2.5.1.patch
```diff
@@ -1,3 +1,27 @@
+diff --git a/node_modules/docusaurus-plugin-typedoc-api/lib/components/MemberSignatureTitle.js b/node_modules/docusaurus-plugin-typedoc-api/lib/components/MemberSignatureTitle.js
```
fnesveda, line 1: Holy shit what kind of magic are we doing here 😄 Wouldn't it be easier to fork the `docusaurus-plugin-typedoc-api` package and publish it under `@apfy-packages`?

B4nan: patch-package is super simple solution, much simpler than forking and publishing to npm :]

fnesveda: This would be a cool topic for an Eat&Learn or Tech Coffee 👍

B4nan: I was actually mentioning that on tech coffee some time ago, together with `yarn link` :] 

You just modify the source inside `node_modules` and run `npx patch-package some-package`, the diff gets generated, and via `postinstall` hook it gets applied automatically (again just a CLI command). It gets tedious when you want to update the dependency, so not really a long-term solution, but super handy for debugging things in production, e.g. adding more logs to some package to test things in production.

fnesveda: Ah, I see it in the history, I was just off that day I think 😄 Shame, looks cool!

PR #110, opened 2023-02-06
fix: Calling user().get() will use '/me' endpoint correctly

jirimoravcik:
The previous code didn't work since we always pass something into `resource_id` from the `.user()` call (the type is `str | None`). This way, we always popped something from `kwargs` and `'me'` could never make it to the URL. 

Now, it will just check whether the value is `None` or not and replace it with `'me'` if it's `None`


fnesveda (approved):
Cool, nice catch!

This `kwargs` popping with defaults is a bit of a mess honestly, maybe all the arguments should be passed explicitly from the callers, and then we wouldn't need to override the constructors at all. But I wanted to have the internal logic of the default IDs and resource paths in the clients that deal with it, rather than in the calling code.

PR #112, opened 2023-02-08
Add `ignore_docs` decorator

fnesveda:
This adds an `ignore_docs` decorator, which does not do anything on its own, it just serves as a mark that we will use in the new documentation rendering to mark methods that we want to ignore there.

I've added it to all the `__init__` methods of all the classes except for `ApifyClient` and `ApifyClientAsync`, because those are the only two classes that we want users to actually create. All the other classes should not be constructed externally.

### DIFF
--- a/src/apify_client/_errors.py
+++ b/src/apify_client/_errors.py
@@ -2,6 +2,8 @@ from typing import Optional
 
 import httpx
 
+from ._utils import ignore_docs
+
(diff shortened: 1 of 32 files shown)


jirimoravcik (approved):

barjin (approved):
Looks great, thanks! ;)

PR #113, opened 2023-02-09
Docs v2

B4nan:


website/docusaurus.config.js
```diff
@@ -7,7 +7,7 @@ const { absoluteUrl } = config;
 
 /** @type {Partial<import('@docusaurus/types').DocusaurusConfig>} */
 module.exports = {
-    title: 'Apify Documentation',
+    title: 'Apify Client for Python',
```
B4nan, line 10: this is the title of the whole docs.apify.com, I think we want it to be the same for all the projects - all the "child docusaurus projects" has it that way on purpose.

unless we want to have the title suffix different, which sounds a bit weird but it might work for our use case

fnesveda: Yeah I wanted to discuss that thing - I think having the project name in the title makes more sense than just "Apify Documentation". Perhaps we could have both there, but it doesn't really make sense to me.

With just "Apify Documentation" there, there's no way to determine from the title which project you're in, because you have the title like "Quick Start | Apify Documentation" for both the client and the SDK, and it gets confusing.

BTW if you open https://docs.apify.com, the title is "Apify Documentation · Apify Documentation | Apify Documentation" 😄

B4nan: Yeah, I know, it is [already reported](https://github.com/apify/apify-docs/issues/519), but you know, there were more serious things to fix :D

Agreed we should have the project-level info in the title, I just want this to be consistent/coordinated.

.editorconfig
```diff
@@ -10,3 +10,6 @@ end_of_line = lf
 
 [Makefile]
 indent_style = tab
+
+[{*.yaml, *.yml}]
+indent_size = 2
```
B4nan, line 15: any reason for this? we do pretty much the opposite in crawlee and others, why not use the same idents across all languages?

fnesveda: We have it like this in all the platform repos, plus it's the default indent used in the GitHub documentation: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

And we had it like this in all the other workflows apart from the docs one, so it was easier to unify it that way.

fnesveda:
@B4nan I'm still planning to do some docs updates for both client and SDK next week, and having the docs in a separate branch where every commit does not trigger integration tests and a new beta release is really handy, so I would keep it separate for a bit more if possible.

I will close the PRs and reopen them when we're ready to merge, so these commits don't notify so many people.

B4nan:
> having the docs in a separate branch where every commit does not trigger integration tests and a new beta release is really handy, so I would keep it separate for a bit more if possible.

We have the CI setup so it does not trigger tests for commits with `docs:` prefix, you should do the same here.

https://github.com/apify/crawlee/blob/master/.github/workflows/test-ci.yml#L13

fnesveda:
Good point, I'll do that too.

PR #116, opened 2023-02-10
Adjust docs publishing workflow

fnesveda:
I've updated the docs publishing workflow a bit, to align better with the other Python workflows, and I simplified it a bit.

It now:
- uses `actions/setup-python@v4` instead of installing Python manually
- puts the command to generate the API reference JSON to the Makefile, so it's easier to test the docs building locally
- updates the GitHub pages publishing steps to use the official actions from GitHub
- adds a `.nojekyll` file so that the GitHub pages are not processed using Jekyll anymore, which speeds up the deploy

I've also updated some docs with fixes to links so I could test the deployment, and it works fine.


.github/workflows/docs.yml
```diff
@@ -1,59 +1,71 @@
...
 
 jobs:
-    build:
-        if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
-        runs-on: ubuntu-latest
-
-        steps:
-            -   uses: actions/checkout@v3
-                with:
-                    token: ${{ secrets.GH_TOKEN }}
```
B4nan, line 18: i am quite sure we need the personal access token, you won't be able to commit from the CI without it if the branch is protected (and maybe even without the branch protection)

fnesveda: Oh, yeah, that makes sense. I will adjust it.

From which user is the `GH_TOKEN`, by the way?

B4nan: It's mine, feel free to replace with yours so you get notified about the docs build failures if you want.

Not sure if we have some GH user we could use to generate those tokens, but it needs to be personal access token, so I believe it needs to be connected to actual GH user account.

fnesveda: We have @apify-service-account exactly for these usecases 🙂 It's already used here in the release workflow for triggering the docker image builds in https://github.com/apify/apify-actor-docker.

I'll make a new token in that user for this and use it here.

barjin (approved):

.github/workflows/docs.yml
fnesveda: Added the token, works fine now even with branch protection.

B4nan:
I guess we should do the same improvements for other docusaurus projects, crawlee included.

fnesveda:
I'll do it for the Python SDK, since there it will be basically a 1:1 copy.

PR #124, opened 2023-05-07
fix: allow newer versions of httpx

martolini:
I've seen this clash with some libraries that want the newest version of httpx. The changelog is here https://github.com/encode/httpx/blob/master/CHANGELOG.md. Would it be possible to allow newer versions of httpx together with this awesome apify client? 🙏

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 
 requires-python = ">=3.8"
 dependencies = [
-    "httpx ~= 0.23.0"
+    "httpx >=0.23.0,<1.0.0"
 ]
 
 [project.optional-dependencies]


fnesveda:
Hi, thanks for the PR! Unfortunately we can't upgrade HTTPX at this point, so I'll have to close it without merging.

We've actually upgraded to HTTPX 0.24.0 on Friday, but then we've had to quickly downgrade back to 0.23.0 because 0.24.0 had [issues parsing IPv6 addresses](https://github.com/encode/httpx/pull/2659) in the `NO_PROXY` environment variable, which was breaking some of our workflows. The bug is already fixed in https://github.com/encode/httpx/pull/2659, so we'll upgrade as soon as a new version of HTTPX is released, which should be [hopefully soon](https://github.com/encode/httpx/issues/2674#issuecomment-1516188989). 🙏 

In the meantime, before we release a new version, if you think that the above mentioned bug won't affect your usecase, you can use the [1.1.0b3](https://pypi.org/project/apify-client/1.1.0b3/) Apify Client version, which depends on HTTPX 0.24.0, but is otherwise the same as the latest version.

By the way, if you don't mind me asking, what are the libraries that you want to use with this client that require HTTPX 0.24? Just curious.

PR #129, opened 2023-07-12
feat: add list_and_lock_head, delete_request_lock, prolong_request_lock, batch_add_requests, batch_delete_requests and list_requests methods

drobnikj:
Python code after a long time, rather check it carefully. 😄

### DIFF
--- a/src/apify_client/clients/resource_clients/request_queue.py
+++ b/src/apify_client/clients/resource_clients/request_queue.py
@@ -74,6 +74,29 @@ class RequestQueueClient(ResourceClient):
 
         return _parse_date_fields(_pluck_data(response.json()))
 
+    def list_and_lock_head(self, *, lock_secs: int, limit: Optional[int] = None) -> Dict:
+        """Retrieve the given number of first requests from the queue and locks them for the given time.
+
+        https://docs.apify.com/api/v2#/reference/request-queues/queue-head-with-locks/get-head-and-lock
+
+        Args:
+            lock_secs (int): How long the second request will be locked for
+            limit (int, optional): How many requests to retrieve
+
+
+        Returns:
+            dict: The desired number of locked requests from the beginning of the queue.
+        """
+        request_params = self._params(lockSecs=lock_secs, limit=limit, clientKey=self.client_key)
+
+        response = self.http_client.call(
+            url=self._url('head/lock'),
+            method='POST',
+            params=request_params,
+        )
+
+        return _parse_date_fields(_pluck_data(response.json()))
+
     def add_request(self, request: Dict, *, forefront: Optional[bool] = None) -> Dict:
         """Add a request to the queue.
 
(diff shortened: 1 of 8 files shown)


jirimoravcik (changes requested):
I'd extend the tests, e.g. lock the requests for 10 seconds, then wait ~11 seconds and check if they were unlocked or something like that. Or maybe we'll do more extensive testing in the SDK so it's ok to have less tests here. We'll see what @fnesveda thinks.

tests/integration/test_request_queue.py
```diff
@@ -0,0 +1,55 @@
...
+
+def random_string(length: int = 10) -> str:
+    return ''.join(random.choice(string.ascii_letters) for _ in range(length))
+
+
+def random_queue_name() -> str:
+    return f'python-client-test-queue-{random_string(5)}'
+
+
+class TestRequestQueueSync:
```
fnesveda, line 15: Some comments and a bit more formatting would be nice here

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -74,6 +74,29 @@ def list_head(self, *, limit: Optional[int] = None) -> Dict:
 
         return _parse_date_fields(_pluck_data(response.json()))
 
+    def list_and_lock_head(self, *, lock_secs: int, limit: Optional[int] = None) -> Dict:
+        """Retrieve the given number of first requests from the queue and locks them for the given time.
```
fnesveda, line 78:
```suggestion
        """Retrieve a given number of unlocked requests from the beginning of the queue and lock them for a given time.
```

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -74,6 +74,29 @@ def list_head(self, *, limit: Optional[int] = None) -> Dict:
...
+
+        https://docs.apify.com/api/v2#/reference/request-queues/queue-head-with-locks/get-head-and-lock
+
+        Args:
+            lock_secs (int): How long the second request will be locked for
```
fnesveda, line 83:
```suggestion
            lock_secs (int): How long the requests will be locked for, in seconds
```

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -170,6 +193,50 @@ def delete_request(self, request_id: str) -> None:
...
         )
 
+    def prolong_request_lock(self, request_id: str, *, lock_secs: int, forefront: Optional[bool] = None) -> Dict:
+        """Prolong the lock on a request.
+
+        https://docs.apify.com/api/v2#/reference/request-queues/request-lock/prolong-request-lock
+
+        Args:
+            request_id (str): ID of the request to prolong the lock
+            lock_secs (int): How long to prolong the lock
```
fnesveda, line 203:
```suggestion
            lock_secs (int): By how much to prolong the lock, in seconds
```

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -170,6 +193,50 @@ def delete_request(self, request_id: str) -> None:
...
+            forefront (bool, optional): Whether to put the request in the beginning or the end of the queue after lock exppires
```
fnesveda, line 204:
```suggestion
            forefront (bool, optional): Whether to put the request in the beginning or the end of the queue after lock expires
```

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -170,6 +193,50 @@ def delete_request(self, request_id: str) -> None:
...
+        return _parse_date_fields(_pluck_data(response.json()))
+
+    def delete_request_lock(self, request_id: str, *, forefront: Optional[bool] = None) -> None:
+        """Delete the lock on a request.
+
+        https://docs.apify.com/api/v2#/reference/request-queues/request-lock/delete-request-lock
+
+        Args:
+            request_id (str): ID of the request to delete the lock
+            forefront (bool, optional): Whether to put the request in the beginning or the end of the queue after lock deleted
```
fnesveda, line 227:
```suggestion
            forefront (bool, optional): Whether to put the request in the beginning or the end of the queue after the lock is deleted
```

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -170,6 +193,50 @@ def delete_request(self, request_id: str) -> None:
...
+
+        response = self.http_client.call(
+            url=self._url(f'requests/{request_id}/lock'),
+            method='PUT',
+            params=request_params,
+        )
+
+        return _parse_date_fields(_pluck_data(response.json()))
+
+    def delete_request_lock(self, request_id: str, *, forefront: Optional[bool] = None) -> None:
```
fnesveda, line 220: We should have called this `unlock_request`, but too late

fnesveda (changes requested):
Nice! I like that you made integration tests for this, I'm sad that we don't have them for the rest of the client 😄 I just had some notes mostly to the docstrings.

And regarding Jirka's point - I'm not sure how complicated the tests need to be here, all the aspects of the platform features should be tested by the platform integration tests, not the client tests, so I think it's fine to just check if the API endpoints work and return what they should, without too many complicated tests.

jirimoravcik (approved):

fnesveda (approved):

fnesveda (changes requested):
I just realized, you should also bump the version and update the CHANGELOG.md

jirimoravcik:
> I just realized, you should also bump the version and update the CHANGELOG.md

how about we automate this going forward?

fnesveda:
> how about we automate this going forward?

I kind of like the versioning flexibility this gives us, plus we weren't making that many changes to these libraries for it to be worth investing the time into. But maybe there is some easy solution for this.

fnesveda (changes requested):
You should bump the version to `1.3.0` in `pyproject.toml`, otherwise the publishing workflow will fail because `1.2.2` already exists.

PR #132, opened 2023-07-25
Start using apify-shared for general consts and utils

### LINKED ISSUE #131: Start using apify-shared for general consts and utils
- In https://github.com/apify/apify-shared-python/issues/1 we introduced `apify-shared` library for sharing general constants and utils across our Python projects. Let's start using it.
  - https://github.com/apify/apify-shared-python
  - https://pypi.org/project/apify-shared/

vdusek:
Closes #131


pyproject.toml
```diff
@@ -23,7 +23,8 @@ classifiers = [
 
 requires-python = ">=3.8"
 dependencies = [
-    "httpx >= 0.24.1"
+    "apify-shared ~= 1.0.0b2",
```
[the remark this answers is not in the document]
fnesveda: Yeah I was writing that too, I wanted to wait until we test that the beta version works, and then publish the stable one, but since the tests have passed here it should be fine

src/apify_client/clients/resource_clients/actor_collection.py
```diff
@@ -1,6 +1,8 @@
 from typing import Any, Dict, List, Optional
 
-from ..._utils import ListPage, _filter_out_none_values_recursively, ignore_docs
+from apify_shared.utils import filter_out_none_values_recursively, ignore_docs
+
+from ..._utils import ListPage
```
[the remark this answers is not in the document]
fnesveda: We decided to keep it in the client because I thought it's kinda client-specific, it's not any shared logic.

But now looking at why we use it in the SDK, that was a mistake, we need it for the local storage emulation, so it's not really client-specific anymore, so we should put it in the shared package.

Question is, where? It's not really a utility, maybe we should have some `apify_shared.models`?

fnesveda (changes requested):
Looks good, just some comments about the consts. And in the end I agree with Jirka's point about `ListPage`, sorry for originally saying that it shouldn't be moved, that was wrong.

src/apify_client/clients/resource_clients/actor_collection.py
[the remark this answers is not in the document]
vdusek: Ok, I'm adding `models.ListPage` to `apify-shared` in https://github.com/apify/apify-shared-python/pull/4.

pyproject.toml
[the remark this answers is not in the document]
vdusek: Ok, I released `apify_shared` `v1.0.0` and use it.

src/apify_client/consts.py
```diff
@@ -1,87 +1,29 @@
...
+    WebhookEventType,
+)
+
+# Trigger DeprecationWarning
+warnings.warn(
+    ('Importing constants from "apify_client.consts" is deprecated and will be removed in the future. '
+     'Please use "apify_shared" library instead.'),
+    category=DeprecationWarning,
+    stacklevel=2,
+)
```
fnesveda, line 29: I think we can't just put a warning when importing this module, we might at some point want to add some non-deprecated client-only constants here (and we do have SDK-only constants in the SDK).

We could solve this by overriding the `__getattr__()` method in this module, as per [PEP 562](https://peps.python.org/pep-0562/).

fnesveda (changes requested):
One more thing about the deprecation warnings in `consts.py`

jirimoravcik (approved):

src/apify_client/consts.py
vdusek: You're right, updated, thanks.

docs/res/sphinx-config/index.rst
```diff
@@ -9,5 +9,3 @@
     :members:
 .. autoclass:: apify_client._utils.ListPage
     :members:
-.. automodule:: apify_client.consts
-    :members:
```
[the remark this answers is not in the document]
vdusek: After a discussion with @fnesveda, we decided that this is gonna be resolved separately in the `docs-v2` branch.

src/apify_client/consts.py
```diff
@@ -1,87 +1,54 @@
...
+    'STRING_ENV_VARS',
+    'STRING_ENV_VARS_TYPE',
+    'ActorEventTypes',
+    'ActorExitCodes',
+    'ActorJobStatus',
+    'ActorSourceType',
+    'ApifyEnvVars',
+    'MetaOrigin',
+    'WebhookEventType',
+]
```
fnesveda, line 40: Maybe we could reexport only the four constants which were here before this PR, to keep it simpler.

src/apify_client/consts.py
```diff
@@ -1,87 +1,54 @@
...
+    if name in DEPRECATED_NAMES:
+        warnings.warn(
+            (
+                f'Importing "{name}" from "apify_client.consts" is deprecated and will be removed in the future. '
+                'Please use "apify_shared" library instead.'
+            ),
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
+        return globals()[f'{name}_']
```
fnesveda, line 53: I think this won't work with any client-only constants we would define in here, it would raise `AttributeError` even if they would be present. Better to have some `else` branch for this.

fnesveda (changes requested):
Few small comments still

src/apify_client/consts.py
vdusek: Actually, I took the structure from [PEP 562](https://peps.python.org/pep-0562/).

And it works, see the following example:
```python
import warnings
from typing import Any

from apify_shared.consts import ActorJobStatus as ActorJobStatus_
from apify_shared.consts import ActorSourceType as ActorSourceType_
from apify_shared.consts import MetaOrigin as MetaOrigin_
from apify_shared.consts import WebhookEventType as WebhookEventType_

DEPRECATED_NAMES = [
    'ActorJobStatus',
    'ActorSourceType',
    'MetaOrigin',
    'WebhookEventType',
]

A = 1


def __getattr__(name: str) -> Any:
    if name in DEPRECATED_NAMES:
        warnings.warn(
            (
                f'Importing "{name}" from "apify_client.consts" is deprecated and will be removed in the future. '
                'Please use "apify_shared" library instead.'
            ),
            category=DeprecationWarning,
            stacklevel=2,
        )
        return globals()[f'{name}_']
    raise AttributeError(f'module {__name__!r} has no attribute {name!r}')
```

Import deprecated const:
```python
>>> from src.apify_client.consts import ActorJobStatus
<stdin>:1: DeprecationWarning: Importing "ActorJobStatus" from "apify_client.consts" is deprecated and will be removed in the future. Please use "apify_shared" library instead.
>>> print(ActorJobStatus)
<enum 'ActorJobStatus'>
```

Import client specific const:
```python
>>> from src.apify_client.consts import A
>>> print(A)
1
```

fnesveda: Ahaaaa, the custom `__getattr__()` is called only when the built-in `object.__getattribute__()` doesn't work, I understand it now. Kinda confusing tbh, but the performance rationale makes sense.

> If an attribute is not found on a module object through the normal lookup (i.e. `object.__getattribute__`), then `__getattr__` is searched in the module `__dict__` before raising an AttributeError. If found, it is called with the attribute name and the result is returned. Looking up a name as a module global will bypass module `__getattr__`. This is intentional, otherwise calling `__getattr__` for builtins will significantly harm performance.

vdusek: An explanatory note was added.

PR #134, opened 2023-07-25
feat(docs): Add indicator for keyword-only arguments, add link to method sources

fnesveda:
The docs were not showing which method arguments are positional and which are keyword-only. This adjusts the docs generation so that there is a `*` in the argument list before the first keyword-only argument, and also that there is a `keyword-only` tag next to those arguments.

And while I was messing with it I adjusted it so that the methods also include a link to their source code.

This might be pretty hard to review code-wise, because it's mostly just package patches, so I think just check if you think the result is OK, or if it should be adjusted. For the future, I vote to fork `docusaurus-plugin-typedoc-api` and use the fork here, so that the changes we make to it are more reviewable.

Once approved, I will do it in the SDK as well.

Screenshot 2023-07-25 at 19 37 15


barjin (approved):
Imo this is good to go, great job with fixing the GitHub links!

One small nit / idea - does the code parser retrieve the default values from initializers (`def f(a = 5, b = 4)`)? For some arguments, this doesn't really make much sense (`api_key`), but for others (`max_retries`, `timeout_secs`) it could be useful to display these 🤔

fnesveda:
It does, I wanted to add it in the next iteration, but maybe I'll do it now since I have it open

fnesveda:
@barjin It's alive! I also fixed the empty "Return" text at the bottom of the constructor:
Screenshot 2023-07-25 at 20 35 26

vdusek (approved):
The result looks great, good job Franto :+1:

PR #135, opened 2023-07-28
Rm warnings simplefilter always

vdusek:


src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -271,7 +271,6 @@ def download_items(
         # We need to override and then restore the warnings filter so that the warning gets printed out,
         # Otherwise it would be silently swallowed
         with warnings.catch_warnings():
-            warnings.simplefilter('always')
```
fnesveda, line 274: Now the `warnings.catch_warnings():` context manager is unnecessary, because we're not changing the warnings filter anymore, we could remove this whole logic and just keep the `warnings.warn(...)` call

fnesveda (changes requested):

src/apify_client/clients/resource_clients/dataset.py
vdusek: Sure, makes sense, I removed the related comment lines as well.

fnesveda (approved):

PR #143, opened 2023-08-25
Improve logging of HTTP requests

fnesveda:
There were some issues in how the requests were logged through our logger, namely:
- the URL in the log was the URL of the resource the request was made to, but not the action on the resource (`RunClient.reboot()` would log URL `https://api.apify.com/v2/actor-runs/RUN_ID`, without the `/reboot` at the end)
- `ApifyClient` and `ApifyClientAsync` had different log messages and details
- the request attempt and method were not logged

This fixes all of that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


vdusek (approved):
LGTM

PR #161, opened 2023-11-14
Migrate from Flake8 & Autopep8 to Ruff

### LINKED ISSUE #133: Remove underscore prefix from objects that are not private
Currently, we use the underscore prefix in objects that are imported from other modules (e.g. all objects in https://github.com/apify/apify-client-python/blob/master/src/apify_client/_utils.py) - are not private.

This was intended to let users of the library know, that these objects are for internal usage only.

However, this does not correspond to the usage of underscore prefixes in the Python world. We should remove these prefixes from the non-private objects.

We can still use the underscore prefix in the module names, to let users know, this module is only for internal usage and should not be imported by library users.

### LINKED ISSUE #160: Migrate to Ruff (linter & formatter)
Ruff is a new extremely fast Python linter written in Rust, which supports many rules from the `flake8` & `pylint` world (700+).

They recently released a formatter, where single quotes are an option :).

vdusek:
### Description

- The vast majority of changes were performed automatically using `ruff format` and `ruff check --fix`.
- Type hints were updated manually.
- To not break anything I tried not to update/fix violations and rather turn off the violation for a particular line (with some exceptions).

### Issue

- Closes #160 
- Closes #133


src/apify_client/clients/resource_clients/schedule_collection.py
```diff
@@ -111,20 +116,20 @@ async def list(
...
         is_enabled: bool,
         is_exclusive: bool,
-        name: Optional[str] = None,
-        actions: Optional[List[Dict]] = None,
-        description: Optional[str] = None,
-        timezone: Optional[str] = None,
-        title: Optional[str] = None,
-    ) -> Dict:
+        name: str | None = None,
+        actions: list[dict] | None = None,  # type: ignore
```
jirimoravcik, line 128: why is the type ignore needed here?

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -101,17 +105,17 @@ def list_items(
...
+            limit=limit,  # type: ignore
+            desc=desc,  # type: ignore
+            clean=clean,  # type: ignore
+            fields=fields,  # type: ignore
+            omit=omit,  # type: ignore
+            unwind=unwind,  # type: ignore
+            skipEmpty=skip_empty,  # type: ignore
+            skipHidden=skip_hidden,  # type: ignore
+            flatten=flatten,  # type: ignore
+            view=view,  # type: ignore
```
jirimoravcik, line 118: too many ignores

jirimoravcik (approved):
I think the PR's good, I added some comments. I know you didn't want to fix them in this PR, but please create an issue and try to improve the parts where you added type or rule ignores, thanks.

src/apify_client/clients/resource_clients/schedule_collection.py
vdusek: solved by allowing Any type

src/apify_client/clients/resource_clients/store_collection.py
```diff
@@ -90,11 +94,11 @@ async def list(
...
-            category=category,
-            username=username,
-            pricingModel=pricing_model,
+            limit=limit,  # type: ignore
+            offset=offset,  # type: ignore
+            search=search,  # type: ignore
+            sortBy=sort_by,  # type: ignore
+            category=category,  # type: ignore
+            username=username,  # type: ignore
+            pricingModel=pricing_model,  # type: ignore
```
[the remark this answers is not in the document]
vdusek: solved by allowing Any type

src/apify_client/clients/resource_clients/schedule_collection.py
vdusek: It is because of the method named `list` in the same module. It breaks `mypy` :confused:. I already created an issue for it - https://github.com/apify/apify-client-python/issues/162.

PR #183, opened 2024-02-22
feat: add monthlyUsage() and limits() methods to UserClient

tobice:
These endpoints have been around for some quite time and now we're exposing them via the Python client.

### DIFF
--- a/src/apify_client/clients/resource_clients/user.py
+++ b/src/apify_client/clients/resource_clients/user.py
@@ -29,7 +29,7 @@ class UserClient(ResourceClient):
         https://docs.apify.com/api/v2#/reference/users
 
         Returns:
-            dict, optional: The retrieved user data, or None if the user does not exist.
+            The retrieved user data, or None if the user does not exist.
         """
         return self._get()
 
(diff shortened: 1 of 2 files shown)


jirimoravcik (approved):

src/apify_client/clients/resource_clients/user.py
```diff
@@ -31,6 +33,55 @@ def get(self: UserClient) -> dict | None:
...
+        """Return monthly usage of the user account.
+
+        This includes a complete usage summary for the current usage cycle, an overall sum, as well as a daily breakdown
+        of usage. It is the same information which is available on the account's Billing page. The information includes
+        use of storage, data transfer, and request queue usage.
+
+        https://docs.apify.com/api/v2/#/reference/users/monthly-usage
+
+        Returns:
+            dict, optional: The retrieved request, or None, if it did not exist.
```
vdusek, line 46: I'd just remove the `dict, optional:` from the docstrings, it's redundant since it's already contained in the type hint. For every endpoint.

vdusek (changes requested):

src/apify_client/clients/resource_clients/user.py
tobice: Good point. Removed in the whole file.

vdusek (approved):
LGTM, thanks

PR #184, opened 2024-02-27
docs: fix CHANGELOG nad pyproject.toml URLs in CONTRIBUTING.md

tobice:
They were leading to Apify Python SDK, which is really confusing 😅


vdusek (approved):
Oh, good catch, thanks :)

PR #191, opened 2024-05-16
chore: Add Python Client examples [docs]

HonzaTuron:
Part of apify/apify-web/issues/3616

Added multiple examples, consulted with @vdusek. 

First three are the same as [JavaScript ones](https://github.com/apify/apify-client-js/pull/548), just reimplemented to Python. Last one is using Python-specific Pandas library for data analysis.


vdusek (changes requested):
The current implementation uses a mix of synchronous and asynchronous interfaces in a somewhat random manner.

I would suggest to either:
- to use only asynchronous interfaces in examples;
- or to provide both sync & async examples for every section.

HonzaTuron:
@vdusek Maybe I could explain differences in the paragraphs above code? I believe this particular examples makes sense with different sync/async approaches

docs/examples.md
```diff
@@ -0,0 +1,141 @@
...
+run_data = actor_client.call(run_input=input_data, timeout_secs=60)
+```
+
+## Manipulating with tasks
+
+To run multiple inputs with the same Actor, most convenient way is to create multiple [tasks](https://docs.apify.com/platform/actors/running/tasks) with different inputs.
+Task input is persisted on Apify platform when task is created.
+
+```python
+from __future__ import annotations
```
vdusek, line 32: This import can be removed.

vdusek (approved):
Ok then, just one minor import-related thing, LGTM otherwise.

docs/examples.md
HonzaTuron:
```suggestion
```

docs/examples.md
```diff
@@ -0,0 +1,140 @@
...
+# Client initialization with the API token
+apify_client = ApifyClient(token='MY_APIFY_TOKEN')
+
+actor_client = client.actor('apify/instagram-hashtag-scraper')
+
+input_data = { 'hashtags': ['rainbow'], 'resultsLimit': 20 }
+
+# Run the Actor and wait for it to finish up to 60 seconds.
+# Input is not persisted for next runs.
+run_data = actor_client.call(run_input=input_data, timeout_secs=60)
```
jbartadev, line 23: We need to try to run the examples all the time, same thing happened to us with the getting started examples in JS client - this will never work because there is no `client` defined in this code

e.g.: this won't work

`actor_client = client.actor('apify/instagram-hashtag-scraper')`

docs/examples.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+## Integration with data analysis libraries (Pandas)
+
+The Apify API client for Python can be easily integrated with data analysis libraries.
+Following example demonstrates how to load items from the last dataset run and pass them to a Pandas DataFrame for further analysis.
+Pandas is a data analysis library that provides data structures and functions to efficiently manipulate large datasets.
+
+```python
+import apify_client
```
jbartadev, line 127: Let's unify the imports across examples - `from apify_client import ApifyClient` or `import apify_client`

jbartadev (changes requested):
Nice! Found some issues

jbartadev (approved):
Seems legit, let's make the API clients great again! 🚀

PR #232, opened 2024-05-29
docs: Added new getting started page 

jbartadev:
I tried to do the same thing as we did for the JS client [here](https://github.com/apify/apify-client-js/pull/542) but for Python.

The goal is to improve onboarding (adaptation) for using our API (in Python in this case)

Snímek obrazovky 2024-05-29 v 17 17 17


docs/index.md
```diff
@@ -0,0 +1,329 @@
...
+
+```python
+from apify_client import ApifyClient
+
+apify_client = ApifyClient(
+    token='MY-APIFY-TOKEN',
+    max_retries=8,
+    min_delay_between_retries_millis=500, # 0.5s
+    timeout_secs=360, # 6 mins
+    )
```
vdusek, line 197: fix indentation

```suggestion
apify_client = ApifyClient(
    token='MY-APIFY-TOKEN',
    max_retries=8,
    min_delay_between_retries_millis=500, # 0.5s
    timeout_secs=360, # 6 mins
)
```

docs/index.md
```diff
@@ -0,0 +1,329 @@
...
+
+    # If there are no more items to fetch, exit the loading
+    if offset + limit >= total:
+        break
+
+    offset += limit
+
+print(f"Overall fetched {len(all_items)} items")
+
+```
```
vdusek, line 304: unify quotes

```suggestion
while True:
    response = dataset_client.list_items(limit=limit, offset=offset)
    items = response.items
    total = response.total
    
    print(f'Fetched {len(items)} items')

    # Merge new items with other already loaded items
    all_items.extend(items)

    # If there are no more items to fetch, exit the loading
    if offset + limit >= total:
        break

    offset += limit

print(f'Overall fetched {len(all_items)} items')
```

vdusek (changes requested):
Nice, thanks!

I have a few code style suggestions to share.

On top of that, one more thing... It's not a requirement, more a personal preference. In many places I'd prefer not to chain so many calls into one command. Breaking it down to more separate commands can make it more understandable for users. Also when you want to transform such long-chained command into async command, you will end up with multiple awaits in one command. Which is not goood.

For example, instead of writing:

```python

apify_client = ApifyClient('MY-APIFY-TOKEN')

# Lists items from the Actor's dataset
dataset_items = apify_client.dataset('daset-id').list_items().items
```

I'd rather use something like:
```python

apify_client = ApifyClient('MY-APIFY-TOKEN')

# Get a dataset client
dataset_client = apify_client.dataset('daset-id')

# Lists items from the Actor's dataset
dataset_items = dataset_client.list_items().items
```

jbartadev:
@vdusek Thanks Vláďo for the suggested improvements! I added a couple of simplifications for the code examples (removed unnecessary chaining). Please let me know if it's ok now

docs/index.md
```diff
@@ -0,0 +1,345 @@
...
+There are some exceptions though, like `list_keys` or `list_head` which paginate differently.
+The results you're looking for are always stored under `items` and you can use the `limit`
+property to get only a subset of results. Other properties can be available depending on the method.
+
+```python
+from apify_client import ApifyClient
+
+apify_client = ApifyClient('MY-APIFY-TOKEN')
+
+# Resource clients accept an ID of the resource.
```
vdusek, line 293: probably the only comment with dot in the end

```suggestion
# Resource clients accept an ID of the resource
```

vdusek (approved):
just one nitpick, otherwise LGTM, thanks :slightly_smiling_face:

HonzaTuron (approved):

TC-MO (approved):
LGTM

PR #266, opened 2024-09-04
docs: fix docs build with `poetry run`

barjin:
Fixes the reference documentation build errors caused by the switch to `poetry` as the sole package manager.

obrazek


janbuchar (approved):
LGTM

Issue #264, opened 2024-09-04
ApifyRequestQueueClient.batch_add_requests cannot handle more than 25 requests


janbuchar:
This is an intentional limitation of the Apify API. Javascript client bypasses this, the python one doesn't.

vdusek:
Once this is done we can revert the https://github.com/apify/apify-sdk-python/pull/261 hotfix.

PR #268, opened 2024-09-13
fix: batch add requests can handle more than 25 requests

vdusek:
### Description

- RQ's batch add requests can handle more than 25 requests.
- It was implemented in the same way as in the TS client.
- Applies for both sync & async versions.
    - I do not use multi-threading for the sync version, since it is out of the scope of this issue. If users want performance, they should go with the client's async version. But if we decide it is worth to implement it, let's open another issue.

### Issues

- Closes: #264

### Testing

It was tested on this code sample (plus the sync alternative):

```python
import asyncio

from apify_client import ApifyClientAsync
from crawlee._utils.crypto import get_random_id

TOKEN = '...'


async def main() -> None:
    apify_client = ApifyClientAsync(token=TOKEN)
    rqs_client = apify_client.request_queues()
    rq = await rqs_client.get_or_create(name='my-rq')
    rq_client = apify_client.request_queue(rq['id'])

    print('Add a single request...')
    result_1 = await rq_client.add_request({'url': 'http://example.com', 'uniqueKey': get_random_id()})
    print(f'result: {result_1}')

    print('Add multiple requests...')
    requests = [
        {
            'url': f'https://example.com/{i}/',
            'uniqueKey': get_random_id(),
        }
        for i in range(110)
    ]
    result_2 = await rq_client.batch_add_requests(requests)
    print(f'result: {result_2}')


if __name__ == '__main__':
    asyncio.run(main())
```

### Checklist

- [x] CI passed

### DIFF
--- a/src/apify_client/clients/resource_clients/request_queue.py
+++ b/src/apify_client/clients/resource_clients/request_queue.py
@@ -5,6 +5,7 @@ import logging
 import math
 from dataclasses import dataclass
 from datetime import timedelta
+from multiprocessing import process
 from typing import Any, TypedDict
 
 from apify_shared.utils import filter_out_none_values_recursively, ignore_docs, parse_date_fields
@@ -586,10 +587,13 @@ class RequestQueueClientAsync(ResourceClientAsync):
         processed_requests = list[dict]()
         unprocessed_requests = list[dict]()
 
-        try:
-            while True:
+        while True:
+            try:
                 batch = await queue.get()
+            except asyncio.CancelledError:
+                break
 
+            try:
                 response = await self.http_client.call(
                     url=self._url('requests/batch'),
                     method='POST',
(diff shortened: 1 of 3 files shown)


vdusek:
I'm not sure what to do with the Redbaron and the `check_async_docstrings` (comparing sync & async docstrings) as they differ (`max_parallel` arg).

janbuchar:
@vdusek regarding Redbaron, that library hasn't had a release in ~5 years, so the issue has been there forever, I suspect. I believe it's safe to ignore that :shrug: 

If we must ensure parity of sync and async docstrings, I guess we can add that argument to the sync version and raise a `NotImplementedError` if someone tries to use it.

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -240,28 +282,78 @@ def delete_request_lock(self: RequestQueueClient, request_id: str, *, forefront:
...
+
+            # Send the batch to the API.
+            response = self.http_client.call(
+                url=self._url('requests/batch'),
+                method='POST',
+                params=request_params,
+                json=list(batch.requests),
+            )
+
+            response_parsed = parse_date_fields(pluck_data(response.json()))
```
janbuchar, line 340: Not that it matters much, but we could put the parsing to the `else` block.

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -540,29 +632,137 @@ async def delete_request_lock(
...
+        """
+        processed_requests = list[dict]()
+        unprocessed_requests = list[dict]()
+
+        while True:
+            # Get the next batch from the queue.
+            try:
+                batch = await queue.get()
+            except asyncio.CancelledError:
+                break
```
janbuchar, line 656: I'm not sure about this - the cancellation may pop up from any `await` statement. But I guess it's fine :smile:

janbuchar (approved):
Looks sane, please consider my comments

src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -1,13 +1,55 @@
...
 from apify_client._utils import catch_not_found_or_throw, pluck_data
 from apify_client.clients.base import ResourceClient, ResourceClientAsync
 
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+logger = logging.getLogger(__name__)
+
+_RQ_MAX_REQUESTS_PER_BATCH = 25
+_MAX_PAYLOAD_SIZE_BYTES = 9 * 1024 * 1024  # 9 MB
```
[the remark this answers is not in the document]
vdusek: They are not.

src/apify_client/clients/resource_clients/request_queue.py
vdusek: sure

PR #283, opened 2024-11-06
chore: update dev configs

vdusek:


.pre-commit-config.yaml
```diff
@@ -1,20 +1,20 @@
 repos:
   - repo: local
     hooks:
-      - id: lint
-        name: Lint codebase
+      - id: lint-check
+        name: Lint check
```
janbuchar, line 5: We'll want to have some regexp-checkable naming convention for code checks soon (so that we can check it before releasing), but I'm not sure this is the one. Maybe we could go for something like "Check: Lint" or "Code check: Lint" instead? I don't know, I'm kinda biased towards prefixes and special characters.

vdusek: This is pre-commit configuration, not GH workflow.

janbuchar (approved):

PR #304, opened 2024-12-03
feat: Add charge method to the run client for "pay per event"

Jkuzz:
Resolves https://github.com/apify/apify-core/issues/18593 by adding the charge endpoint to the run client.

Issue to add docs URL https://github.com/apify/apify-client-python/issues/305

### DIFF
--- a/src/apify_client/clients/resource_clients/run.py
+++ b/src/apify_client/clients/resource_clients/run.py
@@ -244,7 +244,7 @@ class RunClient(ActorJobBaseClient):
         if not event_name:
             raise ValueError('eventName is required for charging an event')
 
-        return self.http_client.call(
+        response = self.http_client.call(
             url=self._url('charge'),
             method='POST',
             headers={
@@ -256,6 +256,7 @@ class RunClient(ActorJobBaseClient):
                 'count': count or 1,
             })
         )
+        return response.json()
 
 
 class RunClientAsync(ActorJobBaseClientAsync):
@@ -488,7 +489,7 @@ class RunClientAsync(ActorJobBaseClientAsync):
         if not event_name:
             raise ValueError('eventName is required for charging an event')
 
-        return await self.http_client.call(
+        response = await self.http_client.call(
             url=self._url('charge'),
             method='POST',
             headers={
@@ -500,3 +501,4 @@ class RunClientAsync(ActorJobBaseClientAsync):
                 'count': count or 1,
             })
         )
+        return response.json()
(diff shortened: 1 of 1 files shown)


src/apify_client/clients/resource_clients/run.py
```diff
@@ -226,6 +228,38 @@ def log(self: RunClient) -> LogClient:
...
+    ) -> dict:
+        """Charge for an event of a Pay-Per-Event Actor run.
+
+        https://docs.apify.com/api/v2#/reference/actor-runs/charge-run/charge-run
+
+        Returns:
+            dict: Status and message of the charge event.
+        """
+        if not event_name:
+            raise ValueError('eventName is required for charging an event')
```
vdusek, line 245:
```suggestion
            raise ValueError('event_name is required for charging an event')
```

src/apify_client/clients/resource_clients/run.py
```diff
@@ -440,3 +474,35 @@ def log(self: RunClientAsync) -> LogClientAsync:
...
```
vdusek, line 492:
```suggestion
            raise ValueError('event_name is required for charging an event')
```

vdusek (commented):
The link to the docs is there, so I believe we can close #305, correct?

vdusek (approved):
LGTM

src/apify_client/clients/resource_clients/run.py
```diff
@@ -226,6 +228,38 @@ def log(self: RunClient) -> LogClient:
...
+            dict: Status and message of the charge event.
+        """
+        if not event_name:
+            raise ValueError('event_name is required for charging an event')
+
+        response = self.http_client.call(
+            url=self._url('charge'),
+            method='POST',
+            headers={
+                'idempotency-key': idempotency_key or f'{self.resource_id}-{event_name}-{int(time.time() * 1000)}',
```
fnesveda, line 251: What if someone tries to charge for two occurences of one event twice in the same millisecond? Then the idempotency key will be the same and one charge will get ignored. Some unique key is needed here, I think a random string would be enough.

There's a `crypto_random_object_id` implementation in [Crawlee for Python](https://github.com/apify/crawlee-python/blob/master/src/crawlee/_utils/crypto.py#L21-L24), perhaps we could move it to [`apify-shared-python`](https://github.com/apify/apify-shared-python/blob/master/src/apify_shared/utils.py) and use it here.

fnesveda (changes requested):
One note about the idempotency key

src/apify_client/clients/resource_clients/run.py
Jkuzz: I took the idempotency key generation from https://github.com/apify-store/contact-info/blob/master/code/src/charging_manager.ts but it could be specific to their actors 🤔 I can move `crypto_random_object_id` 👍

fnesveda (approved):

Issue #306, opened 2024-12-05
Dataset schema validation: add data property to API error object


katacek:
When dataset schema validation fails, js client returns info about it as a part of the 400 error, it would be great if python client can do it as well.

When added, [this](https://github.com/apify/apify-docs/pull/1304#discussion_r1869916423) part of docs can be updated as well (/dataset_schema/validation.md).

Docs PR for dataset schema validation [here](https://github.com/apify/apify-docs/pull/1304)
How it was added in js [here](https://github.com/apify/apify-client-js/pull/559/files)

PR #314, opened 2024-12-13
feat: Add error data to ApifyApiError

Pijukatel:
Add error data to ApifyApiError when available.
Add tests.

Closes: https://github.com/apify/apify-client-python/issues/306


tests/unit/test_client_errors.py
```diff
@@ -0,0 +1,49 @@
...
+_TEST_URL = 'http://example.com'
+_EXPECTED_MESSAGE = 'some_message'
+_EXPECTED_TYPE = 'some_type'
+_EXPECTED_DATA = {
+    'invalidItems': {'0': ["should have required property 'name'"], '1': ["should have required property 'name'"]}
+}
+
+
+@respx.mock
+@pytest.fixture
```
janbuchar, line 20: Is this autoused somehow?

Pijukatel: It is used by 
@pytest.mark.usefixtures('mocked_response')

I would normally just pass the fixture into the args, but then ruff is complaining about unused args :-(, so I thought this somewhat less common way of using fixtures is better than #ignore comment.

janbuchar: I would prefer either fixture(autouse=True) or disabling that rule for tests

pyproject.toml
```diff
@@ -41,7 +41,9 @@ keywords = [
 [tool.poetry.dependencies]
 python = "^3.9"
 apify-shared = ">=1.1.2"
-httpx = ">=0.25.0"
+# TODO: relax the upper bound once the issue is resolved:
+# https://github.com/apify/apify-client-python/issues/313
+httpx = "~0.27.0"
```
vdusek, line 46: If it works with v0.25, we should not make it more strict without a reason. Do we have a reason for that? Otherwise, let's use ">=0.25 < 0.28.0".

Pijukatel: Done

vdusek (approved):
LGTM

PR #324, opened 2025-01-01
feat: add GET: dataset.statistics

MFori:
Part of https://github.com/apify/apify-core/issues/18807 - Implementation of new API endpoint `v2/datasets/{datasetId}/statistics`

### DIFF
--- a/src/apify_client/clients/resource_clients/dataset.py
+++ b/src/apify_client/clients/resource_clients/dataset.py
@@ -7,6 +7,8 @@ from typing import TYPE_CHECKING, Any
 from apify_shared.models import ListPage
 from apify_shared.utils import filter_out_none_values_recursively, ignore_docs
 
+from apify_client._errors import ApifyApiError
+from apify_client._utils import catch_not_found_or_throw, pluck_data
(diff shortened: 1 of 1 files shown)


Pijukatel (commented):
I would suggest to add some basic test for this.

MFori:
> I would suggest to add some basic test for this.

@Pijukatel Sorry I don't write python often, but in js-client we have ready set of tests, for testing that just the right API endpoint is called, but I don't see any similar setup in this repo. The actual endpoint integration tests are in `apify-core` so I don't think it's necessary to test functioning of API endpoint again here.

vdusek:
> @Pijukatel Sorry I don't write python often, but in js-client we have ready set of tests, for testing that just the right API endpoint is called, but I don't see any similar setup in this repo. The actual endpoint integration tests are in apify-core so I don't think it's necessary to test functioning of API endpoint again here.

We currently do not have any testing framework here, so we will have to leave it at that for now.

vdusek (approved):
Looks good :)

Pijukatel (approved):

PR #339, opened 2025-01-27
refactor: rename /openapi-specification -> /openapi.json

danpoletaev:
Recently, we created new endpoints:

`/v2/acts/:actorId/builds/:buildId/openapi-specification`
`/v2/actor-builds/:buildId/openapi-specification`
Than we've received a feedback to rename endpoints to `/openapi.json`.

**This PR renames OpenAPI endpoints from `*/openapi-specification` to `*/openapi.json`*

More context [here](https://apify.slack.com/archives/C01VBUV81UZ/p1737975106009729)

### DIFF
--- a/src/apify_client/clients/resource_clients/build.py
+++ b/src/apify_client/clients/resource_clients/build.py
@@ -43,16 +43,16 @@ class BuildClient(ActorJobBaseClient):
         """
         return self._abort()
 
-    def get_open_api_specification(self) -> dict | None:
-        """Return OpenAPI specification of the Actor's build.
+    def get_open_api_definition(self) -> dict | None:
+        """Return OpenAPI definition of the Actor's build.
 
-        https://docs.apify.com/api/v2/actor-build-openapi-specification-get
+        https://docs.apify.com/api/v2/actor-build-openapi-json-get
 
         Returns:
-            OpenAPI specification of the Actor's build.
+            OpenAPI definition of the Actor's build.
         """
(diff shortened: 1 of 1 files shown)


jirimoravcik (approved):

src/apify_client/clients/resource_clients/build.py
```diff
@@ -43,16 +43,16 @@ def abort(self) -> dict:
         """
         return self._abort()
 
-    def get_open_api_specification(self) -> dict | None:
-        """Return OpenAPI specification of the Actor's build.
+    def get_open_api_definition(self) -> dict | None:
```
B4nan, line 46: isn't it better this way? the endpoint as well as the docs use `openapi` instead of `open-api`.

```suggestion
    def get_openapi_definition(self) -> dict | None:
```

B4nan (approved):

src/apify_client/clients/resource_clients/build.py
danpoletaev: Hm, if we rename it in python, it will also make sense to rename it in js client:

`getOpenApiDefinition()` to `getOpenapiDefinition`

But for me `getOpenApiDefinition()` looks cleaner, what do you think?

B4nan: as mentioned in the edit, it indeed feels weird in the JS version, so lets leave it the way you have it

PR #340, opened 2025-01-28
docs: Move code samples to files and other updates

vdusek:
### Description

- Relocated all code samples to dedicated files for being able to perform static analysis.
- Updated home page example and moved it to a dedicated file.
- Updated `pyproject.toml` configurations to include only directories with source files (`website/` is added there, same as in Crawlee and SDK).
- Revised most of the documentation content.
- All code examples, where it makes sense, are both in sync & async forms.

### Issues

- Closes: #337

### Testing

- All code samples pass checks with Ruff and Mypy.

### Checklist

- [x] CI passed


docs/02_concepts/code/05_retries_async.py
```diff
@@ -0,0 +1,12 @@
+from apify_client import ApifyClientAsync
+
+TOKEN = 'MY-APIFY-TOKEN'
+
+
+async def main() -> None:
+    apify_client = ApifyClientAsync(
+        token=TOKEN,
+        max_retries=8,
+        min_delay_between_retries_millis=500,  # 0.5s
```
Pijukatel, line 10: I know this is just a moved example code, but when the argument units need explanation comment, then probably used name is not sufficiently descriptive. I wonder why not just use SI unit for millisecond: **ms**
(No need to act on this comment in this PR)

website/sidebars.js
```diff
@@ -1,12 +1,52 @@
...
+            label: 'Examples',
+            collapsed: false,
+            items: [
+                {
+                    type: 'autogenerated',
+                    dirName: '03_examples',
+                },
+            ],
         },
+        // {
```
Pijukatel, line 36: What is the point in leaving here commented out code?

vdusek: It will be uncomented and rendered for the potential v2.0.

docs/02_concepts/code/05_retries_async.py
vdusek: Yeah, this is just an example that was moved. We transferred every time-related types to timedelta in SDK & Crawlee. Not here yet.

Pijukatel (approved):

PR #343, opened 2025-01-30
feat: add rate limit statistics

Mantisus:
### Description

- Add `Statistics` for gather HTTP rate limit errors
- Linked Issues is in the sdk, but the implementation should be in the client

### Issues

- https://github.com/apify/apify-sdk-python/issues/318

### Testing

- Add tests for `Statistics`

### DIFF
--- a/src/apify_client/_statistics.py
+++ b/src/apify_client/_statistics.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from dataclasses import dataclass, field
 
 
@@ -11,7 +12,7 @@ class Statistics:
     requests: int = 0
     """Total number of HTTP requests sent, including retries."""
 
-    rate_limit_errors: list[int] = field(default_factory=list)
+    rate_limit_errors: defaultdict[int, int] = field(default_factory=lambda: defaultdict(int))
     """List tracking which retry attempts encountered rate limit (429) errors."""
 
     def add_rate_limit_error(self, attempt: int) -> None:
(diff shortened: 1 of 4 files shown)


src/apify_client/_statistics.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass, field
+
+
+@dataclass
+class Statistics:
+    """Statistics about API client usage and rate limit errors."""
+
+    calls: int = 0
+    requests: int = 0
+    rate_limit_errors: list[int] = field(default_factory=list)
```
vdusek, line 10: Maybe we could utilize [`defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict) and get rid of `_ensure_list_capacity`?

src/apify_client/_statistics.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def add_rate_limit_error(self, attempt: int) -> None:
+        """Add rate limit error for specific attempt.
+
+        Args:
+            attempt: The attempt number (1-based indexing)
```
vdusek, line 16:
```suggestion
            attempt: The attempt number (1-based indexing).
```

src/apify_client/_statistics.py
vdusek, line 10: Could you please describe these fields?

tests/unit/test_statistics.py
```diff
@@ -0,0 +1,44 @@
...
+        ([1, 5, 3], [1, 0, 1, 0, 1]),
+        ([2, 1, 2, 1, 5, 2, 1], [3, 3, 0, 0, 1]),
+    ],
+)
+def test_add_rate_limit_error(attempts: list[int], expected_errors: list[int]) -> None:
+    """Test that add_rate_limit_error correctly tracks errors for different attempt sequences."""
+    stats = Statistics()
+    for attempt in attempts:
+        stats.add_rate_limit_error(attempt)
+    assert stats.rate_limit_errors == expected_errors
```
vdusek, line 22: Could you utilize pytest param with IDs for better readability of test cases?

src/apify_client/_statistics.py
Mantisus: I'm not sure I see a cleaner solution with `defaultdict` that would give us the same result.

This is if we want to keep the TS-like data format as an array. If we don't have such a need, then we can really go with `dict`

vdusek (approved):
LGTM

PR #383, opened 2025-04-08
Add `DynamicTimeoutFunction`

Pijukatel:
Add `DynamicTimeoutFunction` to enable custom timeouts for individual requests.

This allows users of `ApifyClientAsync` and `ApifyClient` to pass optional argument `get_dynamic_timeout` which is a function that can get suitable timeout for http requests to Apify API. Clients will use this function to set timeout for the requests. 
Clients behave the same as before if the optional argument is not used.

Enables fix of: https://github.com/apify/crawlee-python/issues/1132

### DIFF
--- a/src/apify_client/_dynamic_timeout.py
+++ b/src/apify_client/_dynamic_timeout.py
@@ -1,0 +1,13 @@
+from __future__ import annotations
+
+from collections.abc import AsyncIterable, Iterable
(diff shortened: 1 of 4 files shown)


Pijukatel:
Close din favor of: https://github.com/apify/apify-client-python/pull/384

PR #389, opened 2025-04-22
fix: default_build() returns BuildClient

danpoletaev:
This PR updates the `Actor.default_build()` method to return a BuildClient instance instead of a raw response object.


vdusek (approved):
LGTM

Issue #390, opened 2025-04-23
Refactor retries in `RequestQueueClient(Async).batch_add_requests`


Pijukatel:
Currently the `batch_add_requests` method of `RequestQueueClient(Async)` is doing [some retries](https://github.com/apify/apify-client-python/blob/master/src/apify_client/clients/resource_clients/request_queue.py#L698) based on the status code response, this seems to duplicate retries already happening in `HttpClient`, that [is already retrying the requests as well.](https://github.com/apify/apify-client-python/blob/master/src/apify_client/_http_client.py#L218)

The retries in `RequestQueueClient(Async)` seems unnecessary.

PR #391, opened 2025-04-24
refactor: Remove redundant retries from `batch_add_requests`

Pijukatel:
### Description

Remove retries from `batch_add_requests` (such retries already exist on the http client level).
Add deprecation warnings when retires related arguments are used.
Align sync and async version to not ignore exceptions from the http client. 
Add tests.

### Issues

- Closes: #390


vdusek (commented):
Breaking change? Are we planning to release client v2? 🙂
We'll probably need to make this backward compatible for now, with a follow-up issue to clean things up for v2.0.

vdusek (approved):
thanks

PR #394, opened 2025-04-28
feat: add support for general resource access

tobice:
We plan to allow users to configure through API & clients how their storages and runs can be accessed — whether just the ID / name is sufficient to access the storage. This is controlled through the `general_access` resource property, which overrides account setting for the resource.

Full context in https://github.com/apify/apify-core/issues/19012.

This PR updates the client so that it can be used to get and update general access on datasets, key-value stores, request queues, and runs.

```python
client = ApifyClient(token=API_TOKEN)

# Set general access to ANYONE_WITH_ID_CAN_READ which overrides the account setting.
client.dataset(dataset_id).update(
    general_access=StorageGeneralAccess.ANYONE_WITH_ID_CAN_READ
)
```


vdusek (changes requested):
Could you please use the new version of `uv` and regenerate the lock file (you're removing the `upload_time` fields)? :slightly_smiling_face:

danpoletaev (approved):

vdusek (approved):
LGTM

PR #397, opened 2025-05-02
feat: add list kv store keys by collection or prefix

MFori:
Add `collection` and `prefix` query params for listing key-value store keys based on https://github.com/apify/actor-whitepaper/blob/master/pages/KEY_VALUE_STORE_SCHEMA.md#api-implications

Api docs: https://github.com/apify/apify-docs/pull/1564
Api impl: https://github.com/apify/apify-core/pull/20726


vdusek (approved):
LGTM

Issue #402, opened 2025-05-13
Redirect logs from other actors


Pijukatel:
Add convenient way to redirect logs from actor run to another logger.

This is one part of larger feature request in JS version of the client: https://github.com/apify/apify-client-js/issues/632
(Will be first implemented in Python and then ported to JS version)

PR #403, opened 2025-05-13
feat: Add redirected actor logs

Pijukatel:
### Description

- Add convenience methods and arguments for redirecting streamed actor log into another logger.
- This changes the default behavior in non breaking way. By default the logs will be redirected now.
- Update log endpoints calls to accept `raw` parameter.

#### Starting new actor run through call

3 different use-cases:

```python
...
async with Actor:
    # Default logger
    await Actor.new_client().actor(actor_id="...").call(timeout_secs=5)

    # No logger
    await Actor.new_client().actor(actor_id="...").call(timeout_secs=5, logger = None)

    # Custom logger
    custom_logger = logging.getLogger("apify.kustom_logga")
    handler = logging.StreamHandler()
    custom_logger.addHandler(handler)
    custom_logger.propagate = False

    await Actor.new_client().actor(actor_id="SbjD4JEucMevUdQAH").call(timeout_secs=5, logger = custom_logger)
```

#### Attaching to already running actor, redirecting all logs from actor start

(This is the technique to be used when starting actor through `actor.start` as well)

```python
...
async with Actor:
    ...
    # With context manager
    async with await Actor.new_client().run("some_id").get_streamed_log():
        await asyncio.sleep(5)
        # Logging will stop out of context

     # With direct call (lifecycle of logging has to be manually handled)
    streamed_log = await Actor.new_client().run("some_id").get_streamed_log()
    streamed_log.start()
    await asyncio.sleep(5)
    streamed_log.close()
```

#### Attaching to already running actor, redirecting only new logs
(This can be useful when communicating with some long-running standby actor)

```python
...
async with Actor:
    # With context manager
    async with await Actor.new_client().run("some_id").get_streamed_log(from_start=False):
        await asyncio.sleep(5)
        # Logging will stop out of context
```

#### Example actor with all the logging options shown:
https://console.apify.com/actors/m7ZFYX42SCRFOL0JW

### Issues

- Closes: #402


Pijukatel:
This is a minimal draft version with only async client updated so far to have early review on the concept (sync client will be done later).

Such change can be later propagated to Actor class in sdk.

Example usage shown on the screenshot:

1. Default - redirect logger (will be new default behavior), user do not need to do anything
2. No redirect logger (same as current behavior, but have to be selected by explicitly passing `None` to logger argument)
3. Custom logger defined by the user

image

Redirected messages have estimated logging level, so in most cases you can figure out what level it was in the original log and set it accordingly in this log.

Not in the scope of first version: Respecting the coloring of the original logs -> This information is not available in the logs returned from the api call, so we could try to recreate the coloring during runtime by inspecting all the messages and injecting color markers where suitable, but that could be very computational heavy, so not sure if it is worth it. It would be nice if api could return the colored original, but that would require saving the logs twice, which is also probably not worth it.

Pijukatel:
The redirected logs are now colored
image

Thanks @barjin for exposing the secret parameter!

Pijukatel:
Original log colors are now respected:
image

Thanks a lot @barjin for letting me know about the secret parameter of the log endpoint!

src/apify_client/clients/resource_clients/log.py
```diff
@@ -175,3 +202,173 @@ async def stream(self) -> AsyncIterator[httpx.Response | None]:
...
+        all_parts = re.split(self._split_marker, ''.join(self._stream_buffer))[1:]  # First split is empty string
+        if include_last_part:
+            message_markers = all_parts[0::2]
+            message_contents = all_parts[1::2]
+            self._stream_buffer = []
+        else:
+            message_markers = all_parts[0:-2:2]
+            message_contents = all_parts[1:-2:2]
+            # The last two parts (marker and message) are possibly not complete and will be left in the buffer
+            self._stream_buffer = all_parts[-2:]
```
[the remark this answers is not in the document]
Pijukatel: Negative indexing works fine even if the list is smaller.

src/apify_client/clients/resource_clients/actor.py
```diff
@@ -313,6 +315,9 @@ def call(
                 a webhook set up for the Actor, you do not have to add it again here.
             wait_secs: The maximum number of seconds the server waits for the run to finish. If not provided,
                 waits indefinitely.
+            logger: Loger used to redirect logs from the Actor run. By default, it is set to "default" which means that
```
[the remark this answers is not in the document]
Pijukatel: Done

Pijukatel:
No need to do deep review @MQ37. I think you might be using this functionality soon. Could you please take a look at the use-cases and see if it is convenient to use from your point of view and if there is missing something?
(Timeouts and status messages will be done in separate PR)

src/apify_client/clients/resource_clients/log.py
```diff
@@ -175,3 +202,177 @@ async def stream(self) -> AsyncIterator[httpx.Response | None]:
...
+        """
+        self._to_logger = to_logger
+        if self._force_propagate:
+            to_logger.propagate = True
+        self._stream_buffer = list[str]()
+        self._split_marker = re.compile(r'(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z)')
+        self._relevancy_time_limit: datetime | None = None if from_start else datetime.now(tz=timezone.utc)
+
+    def _process_new_data(self, data: bytes) -> None:
+        new_chunk = data.decode('utf-8')
```
barjin, line 239: A small binary nit:

image

If the `aiter_bytes()` chunk boundary splits a UTF-8 multibyte character (e.g. `í = 0xC3 ][ 0xAD`), you end up with an exception. I'm not sure what the best solution here is - maybe keep the buffer as a binary string, scan for `\n` (`0x0D`) byte and only decode (and print) once you've read from one `\n` to the next one?

src/apify_client/clients/resource_clients/run.py
```diff
@@ -515,6 +548,32 @@ def log(self) -> LogClientAsync:
             **self._sub_resource_init_options(resource_path='log'),
         )
 
+    async def get_streamed_log(
```
janbuchar, line 551: I'm confused with the intended usage here. Are you supposed to call this method and then pass the result into an `actor.call(logger=...)` call?

janbuchar (commented):
I haven't managed to read the whole thing yet, but here are some comments

MQ37 (approved):
Just checked the example usage and from UX viewpoint this looks good, thank you :+1:

src/apify_client/clients/resource_clients/log.py
Pijukatel: Good point. I will keep bytes in the buffer and postpone the decoding to be done only on parts bounded by the marker, which should be always complete and safe to decode.
I added one such `\xc3\xa1` character into the test and split it into two chunks.
https://github.com/apify/apify-client-python/pull/403/commits/cba571fdf2b6263328a2cfd135862eabe6b01e4d

pyproject.toml
```diff
@@ -25,6 +25,7 @@ classifiers = [
 keywords = ["apify", "api", "client", "automation", "crawling", "scraping"]
 dependencies = [
     "apify-shared>=1.4.1",
+    "colorama~=0.4.0",
```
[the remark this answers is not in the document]
Pijukatel: Tiny. Less than 30 kB, https://pypi.org/project/colorama/#files
(It is already part of sdk and crawlee, so for those downstream repos it will not be even any change)

src/apify_client/clients/resource_clients/actor.py
```diff
@@ -313,6 +315,9 @@ def call(
                 a webhook set up for the Actor, you do not have to add it again here.
             wait_secs: The maximum number of seconds the server waits for the run to finish. If not provided,
                 waits indefinitely.
+            logger: Logger used to redirect logs from the Actor run. By default, it is set to "default" which means that
+                the default logger will be created and used. Setting `None` will disable any log propagation. Passing
+                custom logger will redirect logs to the provided logger.
```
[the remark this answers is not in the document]
Pijukatel: I tried to rephrase it like this:

> .... Using "default" literal means that a predefined default logger will be used. ...

src/apify_client/clients/resource_clients/log.py
```diff
@@ -11,8 +18,10 @@
 
 if TYPE_CHECKING:
     from collections.abc import AsyncIterator, Iterator
+    from types import TracebackType
 
     import httpx
+    from mypy.types import Self
```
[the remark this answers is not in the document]
Pijukatel: Ok, chaged.

src/apify_client/clients/resource_clients/run.py
Pijukatel: No, actor call is one group of use-cases that deal with you starting the other actor and then `get_streamed_log` is called internally.

Using this `get_streamed_log` method directly has different use-case. It is when you want to stream log from already started actor (for example long-running actor in standby)

```
async with await Actor.new_client().run("some_id").get_streamed_log(actor_name="CoolActor", from_start=False):
    # Do stuff while logs from other actor are redirceted
```

src/apify_client/clients/resource_clients/log.py
```diff
@@ -175,3 +202,179 @@ async def stream(self) -> AsyncIterator[httpx.Response | None]:
         finally:
             if response:
                 await response.aclose()
+
+
+class StreamedLog:
+    """Utility class for streaming logs from another actor.
```
vdusek, line 208: Please use capitalized Actor (everywhere, there are many occurrences).

```suggestion
    """Utility class for streaming logs from another Actor.
```

vdusek (changes requested):
Does the Actor call log propagation work recursively? (Actor calls an Actor who calls another Actor...)

Pijukatel:
> Does the Actor call log propagation work recursively? (Actor calls an Actor who calls another Actor...)

That is a really good point. It does if the actors were build with version that includes this change. On the other hand the line breaking does not work as expected anymore as the time marker is present for each log (so with two level of log redirection it will be there 3 times and thus there will be extra split)

See example log from this recursive actor:
image

Removing timestamp in redirected logs is probably also not OK as those timestamps are important especially for logs that happened before the streaming started. 
That is something to figure out before merging this.

Pijukatel:
> Does the Actor call log propagation work recursively? (Actor calls an Actor who calls another Actor...)

I updated the split pattern and test data to properly respect the real data and now it works. See example of 4 level recursive actor and it's logs:

image

uv.lock
```diff
@@ -70,24 +73,25 @@ dev = [
...
     { name = "ruff", specifier = "~=0.11.0" },
     { name = "setuptools" },
+    { name = "types-colorama", specifier = "~=0.4.15.20240106" },
 ]
 
 [[package]]
 name = "apify-shared"
 version = "1.4.1"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/b2/a6/c8e2fa0b3bdc479d3ecde778e2381af199f910cf7c8baa3c207bcfe26e47/apify_shared-1.4.1.tar.gz", hash = "sha256:16e617c840fd27bf38d980f079c0b867c7378f68c7006b3d5a7d530d43930507", size = 13871, upload-time = "2025-04-28T12:20:01.113Z" }
```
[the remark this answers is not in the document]
Pijukatel: Updated to `uv==0.7.4` and regenerated the lock file.

src/apify_client/clients/resource_clients/log.py
Pijukatel: Done

tests/unit/test_logging.py
```diff
@@ -0,0 +1,334 @@
...
+    for expected_message_and_level, record in zip(_EXPECTED_MESSAGES_AND_LEVELS, caplog.records):
+        assert expected_message_and_level[0] == record.message
+        assert expected_message_and_level[1] == record.levelno
+
+
+@respx.mock
+def test_actor_call_redirect_logs_to_custom_logger_sync(
+    caplog: LogCaptureFixture,
+    mock_api_sync: None,  # noqa: ARG001, fixture
+    propagate_stream_logs: None,  # noqa: ARG001, fixture
```
vdusek, line 321: Could we do it better here? Maybe [`pytest-mark-usefixtures`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-usefixtures).

Pijukatel: @janbuchar  I could not find the comment, but I remember we were already discussing it some time ago in another review and you were in favor of # noqa.

I have no problem using this fixture and I am also OK with the # noqa, so if you agree with @vdusek on the preferred way in our code base, I will just use it.

vdusek (approved):
LGTM

tests/unit/test_logging.py
Pijukatel: no change then :-)

PR #408, opened 2025-05-20
feat: Add `unlock_requests` method to RequestQueue clients

drobnikj:
Introduce the `unlock_requests` method to both sync and async RequestQueue clients, allowing users to unlock all requests locked by the same clientKey or Actor run.


vdusek (approved):
LGTM

PR #419, opened 2025-06-13
feat: being able to add headers template in webhooks created dynamically

gaelloyoly:
Add usage of headers template
https://docs.apify.com/platform/integrations/webhooks/actions#headers-template

### DIFF
--- a/src/apify_client/_utils.py
+++ b/src/apify_client/_utils.py
@@ -128,6 +128,8 @@ def encode_webhook_list_to_base64(webhooks: list[dict]) -> str:
         }
         if 'payload_template' in webhook:
             webhook_representation['payloadTemplate'] = webhook['payload_template']
+        if 'headers_template' in webhook:
+            webhook_representation['headersTemplate'] = webhook['headers_template']
         data.append(webhook_representation)
 
     return base64.b64encode(json.dumps(data).encode('utf-8')).decode('ascii')


vdusek (approved):
LGTM

Pijukatel:
Could you please describe your top level call (Which client did you use, or did you call some `Actor` method from Apify SDK?) Because I see, that there is a function that is already responsible for conversion from underscore names: https://github.com/apify/apify-client-python/blob/master/src/apify_client/clients/resource_clients/webhook.py#L24 
, but most likely there is some path that is not using this function, so that is why I want to understand how you called it.

Pijukatel:
> Yes, it's the "start" method of the actor client. I'm doing something like this :
> 
> ```python
> from apify_client import ApifyClient
> client = ApifyClient("APIFY_API_KEY")
> client.actor("actor_id").start(
>     run_input={"url": "input_url"},
>     webhooks=[{
>         "event_types": ["ACTOR.RUN.SUCCEEDED"],
>         "request_url": "https://example.com/success",
>         "payload_template": "{{resource}}",
>         "headers_template": '{"x-custom-signature": "1234567890"}',
>     }],
> )
> ```

Thank you for the PR and the clarification. I created issue to refactor the part of the code responsible for handling the Webhooks inputs and in the meantime let's merge your PR.

Pijukatel (approved):

PR #422, opened 2025-06-16
feat: allow sorting of Actors collection

protoss70:
Added new sorting param sortBy for the new feature described in this [Issue](https://app.zenhub.com/workspaces/platform-team-5f6454160d9f82000fa6733f/issues/gh/apify/apify-core/20997)

I will wait until [Apify-Core PR](https://github.com/apify/apify-core/pull/21566) is deployed and then will merge this.

One note, I am not sure if I added the sortBy parameter correctly. Example code below:
`return await self._list(my=my, limit=limit, offset=offset, desc=desc, sortBy=sort_by)`
I tried to follow similar examples with a camel case api parameter. Please let me know if this is not the correct syntax.

### DIFF
--- a/src/apify_client/clients/resource_clients/actor_collection.py
+++ b/src/apify_client/clients/resource_clients/actor_collection.py
@@ -152,6 +152,7 @@ class ActorCollectionClientAsync(ResourceCollectionClientAsync):
         limit: int | None = None,
         offset: int | None = None,
         desc: bool | None = None,
+        sort_by: Literal["createdAt", "lastRunStartedAt"] | None = "createdAt",
     ) -> ListPage[dict]:
         """List the Actors the user has created or used.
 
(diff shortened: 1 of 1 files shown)


src/apify_client/clients/resource_clients/actor_collection.py
```diff
@@ -36,11 +37,12 @@ def list(
             limit: How many Actors to list.
             offset: What Actor to include as first when retrieving the list.
             desc: Whether to sort the Actors in descending order based on their creation date.
+            sort_by: Field to sort the results by. Allowed values are "createdAt" and "lastRunStartedAt".
```
vdusek, line 40: This is redundant, and not good for maintainability.

```suggestion
            sort_by: Field to sort the results by.
```

vdusek (changes requested):
Looks good :slightly_smiling_face: , just one note regarding the docstrings.

PR #432, opened 2025-07-02
chore: Bump minimum Python version to 3.10

vdusek:


Pijukatel (commented):
You will have to update the integration test yaml, seems like I missed that one when parametrizing the python yaml files by python version

https://github.com/apify/workflows/pull/187/files

.github/workflows/run_code_checks.yaml
```diff
@@ -31,4 +37,6 @@ jobs:
     name: Integration tests
     needs: [lint_check, type_check, unit_tests]
     uses: apify/workflows/.github/workflows/python_integration_tests.yaml@main
+    with:
+      python-version: '["3.10", "3.11", "3.12", "3.13"]'
```
Pijukatel, line 41: Previously, we were running just latest and newest 3.9 and 3.13 versions for integration tests, should we retain this logic and run 3.10 and 3.13 only? 

I am fine with both. Just want to make sure this is intentional.

Pijukatel (approved):

.github/workflows/run_code_checks.yaml
vdusek: Yeah, this is intentional for the SDK, because there integration tests run for ~10mins. Here it is only for a few secs, so it's okay to run them on all versions. Thanks.

PR #444, opened 2025-07-25
chore: Move `ListPage` from `apify_shared` 

Pijukatel:
### Description

`ListPage` is used only here. This will make it possible to render it in docs.

### DIFF
--- a/src/apify_client/clients/base/resource_collection_client.py
+++ b/src/apify_client/clients/base/resource_collection_client.py
@@ -1,12 +1,40 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Generic, TypeVar
 
-from apify_shared.models import ListPage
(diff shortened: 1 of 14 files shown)


src/apify_client/clients/base/resource_collection_client.py
```diff
@@ -1,13 +1,41 @@
...
+    #: int: Count of the returned objects on this page
+    count: int
+    #: int: The limit on the number of returned objects offset specified in the API call
+    offset: int
+    #: int: The offset of the first object specified in the API call
+    limit: int
+    #: int: Total number of objects matching the API call criteria
+    total: int
+    #: bool: Whether the listing is descending or not
+    desc: bool
```
vdusek, line 27: Since this is gonna be rendered now, can we use standard docstrings for fields?

vdusek (approved):

PR #453, opened 2025-08-01
feat: add dataset.create_items_public_url and key_value_store.create_keys_public_url

danpoletaev:
When storage resources (Datasets or Key-Value Stores) are set to Restricted, accessing or sharing their data externally becomes difficult due to limited permissions. This PR introduces functionality to generate signed URLs that allow controlled external access to these resources without adding token to the request.

This PR introduces methods to generate signed URLs for Dataset items and Key-Value Store records:
1. **Datasets**
`dataset(:datasetId).create_items_public_url(options, expires_in_millis)`
→ Returns a signed URL like:
`/v2/datasets/:datasetId/items?signature=xxx`

2. Key-Value Stores
`key_value_store(:storeId).create_keys_public_url(options, expires_in_millis)`
→ Returns a signed URL like:
`/v2/key-value-stores/:storeId/keys?signature=xxx`

🕒 Expiration:
The `expires_in_millis` parameter defines how long the signature is valid.
- If provided, the URL will expire after the specified time.
- If omitted, the URL will never expire.

Note 1: The signature is included only if the token has WRITE access to the storage. Otherwise, an unsigned URL is returned.

P.S. We're not yet exposing `urlSigningSecretKey` for datasets, it will be released after [PR](https://github.com/apify/apify-core/pull/22173) is merged.

[More context here](https://www.notion.so/apify/Signed-Dataset-Items-KV-store-record-URLs-224f39950a2280158a6bd82bc2e2ebb5?source=copy_link)

Same PR in JS https://github.com/apify/apify-client-js/pull/720


src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -571,6 +572,67 @@ def get_statistics(self) -> dict | None:
...
+            fields=fields,
+            omit=omit,
+            unwind=unwind,
+            skipEmpty=skip_empty,
+            skipHidden=skip_hidden,
+            flatten=flatten,
+            view=view,
+        )
+
+        if dataset and 'urlSigningSecretKey' in dataset:
```
tobice, line 621: Hmmm starting here the rest of the function seems identical to the async variant. Now sure if there is a nice way to reuse the code? I guess this is how we do it here in the client? 🤔

src/apify_client/_utils.py
```diff
@@ -149,3 +152,59 @@ def encode_key_value_store_record_value(value: Any, content_type: str | None = N
...
+    HMAC signature is truncated to 30 characters to make it shorter.
+
+    Args:
+        secret_key (str): Secret key used for signing signatures
+        message (str): Message to be signed
+
+    Returns:
+        str: Base62 encoded signature
+    """
+    signature = hmac.new(secret_key.encode('utf-8'), message.encode('utf-8'), hashlib.sha256).hexdigest()[:30]
```
tobice, line 186: Just curious: Why the `encode` all over the place? 🤔

tests/integration/key_value_store.py
```diff
@@ -0,0 +1,106 @@
...
+
+import httpx
+from integration_test_utils import random_resource_name
+
+if TYPE_CHECKING:
+    from apify_client import ApifyClient, ApifyClientAsync
+
+
+class TestKeyValueStoreSync:
+    def test_key_value_store_should_create_public_keys_expiring_url_with_params(
```
tobice, line 15:
```suggestion
    def test_key_value_store_should_create_expiring_keys_public_url_with_params(
```

tests/integration/key_value_store.py
```diff
@@ -0,0 +1,106 @@
...
+        httpx_client = httpx.Client()
+        response = httpx_client.get(keys_public_url, timeout=5)
+        assert response.status_code == 200
+
+        time.sleep(3)
+
+        # Assert that the request is now forbidden (expired signature)
+        httpx_client = httpx.Client()
+        response_after_expiry = httpx_client.get(keys_public_url, timeout=5)
+        assert response_after_expiry.status_code == 403
```
tobice, line 38: So these tests are actually executed against real APIs? No mocking? Nice.

tobice (approved):
Nothing surprising here, so here is my approve 👍 I'll leave the Python stuff to Python experts.

tests/integration/key_value_store.py
danpoletaev: Yeah, i was also surprised 😅 
And we can actually test it against multi-staging as well. Would love to see this also in JS client

src/apify_client/clients/resource_clients/dataset.py
janbuchar: We sure do repeat stuff here, but no need to change that now. Perhaps one day we can generate the whole client lib from an OpenAPI spec. Until then, this will have to do

src/apify_client/_utils.py
```diff
@@ -149,3 +152,59 @@ def encode_key_value_store_record_value(value: Any, content_type: str | None = N
         value = json.dumps(value, ensure_ascii=False, indent=2, allow_nan=False, default=str).encode('utf-8')
 
     return (value, content_type)
+
+
+# TODO: will be removed once create_hmac_signature is moved to apify_shared.utils
+# https://github.com/apify/apify-shared-python/pull/44
+CHARSET = string.digits + string.ascii_letters
```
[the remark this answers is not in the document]
danpoletaev: Apify shared python was released and new version was installed 👍

src/apify_client/clients/resource_clients/dataset.py
```diff
@@ -560,6 +561,67 @@ def get_statistics(self) -> dict | None:
...
+        clean: bool | None = None,
+        desc: bool | None = None,
+        fields: list[str] | None = None,
+        omit: list[str] | None = None,
+        unwind: list[str] | None = None,
+        skip_empty: bool | None = None,
+        skip_hidden: bool | None = None,
+        flatten: list[str] | None = None,
+        view: str | None = None,
+        expires_in_millis: int | None = None,
```
janbuchar, line 578: Now I noticed, in the user-facing code in the client, we usually use seconds, not milliseconds

danpoletaev: Changed to `expires_in_seconds`

danpoletaev: Changed it ✅

janbuchar (approved):

PR #454, opened 2025-08-06
chore: replace `respx` mocks with `pytest-httpserver` in tests

Mantisus:
- Replace `respx` mocks with `pytest-httpserver` in tests


Mantisus:
I chose `pytest-httpserver` as a testing tool. This is because the tests are too specific and do not allow the use of `httpbin` without breaking the test logic. At the same time, there are not enough tests to use `uvicorn` with complex endpoint logic.

I haven't deleted `respx` yet, as the timeout tests are quite specific to `httpx`. They should be considered separately after switching to `impit`.

Pijukatel (approved):
I think it is good, but I am curious about the impit and those `test_specific_timeouts_...` tests

PR #455, opened 2025-08-06
feat: extend status parameter to an array of possible statuses 

JanHranicky:
The `status` parameter in the Apify Core API has been extended on the following endpoints:

- `/v2/acts/:actorId/runs`
- `/v2/actor-runs`
- `/v2/actor-tasks/:actorTaskId/runs`

It is now a comma-separated string of possible statuses.

This PR updates the python client accordingly by modifying the `RunCollectionClient.list()` and `RunCollectionClientAsync.list()` methods to reflect the API change. Test case `TestRunCollectionSync` was added to cover  the change


protoss70 (approved):
LGTM :)

PR #456, opened 2025-08-07
refactor!: Replace `httpx` HTTP client with `impit`

Mantisus:
- Replace `httpx` http-clients with `impit`


Mantisus:
The errors in the tests are caused by issues with `impit.Client`, `Response`, and multithreading.

I expect them to be resolved with https://github.com/apify/impit/pull/234

When testing locally on a build with this fix, the errors do not reproduce.

vdusek:
We consider this as BC, and I don't think it is a feature, so I  updated the title type to `refactor!`.

vdusek (changes requested):
Could you also please describe the BCs in the upgrading guide? Similar to what we have in Crawlee and SDK, see e.g. https://crawlee.dev/python/docs/upgrading/upgrading-to-v1.

vdusek (approved):
LGTM!

PR #466, opened 2025-08-13
refactor!: Delete the re-imports in the `consts.py`

Mantisus:
- Delete the re-imports in the `consts.py`. I believe that the temporary period for this has ended, and it is no longer used anywhere.

### Issues

- Closes: #163


vdusek (approved):
LGTM

Issue #158, opened 2023-10-25
Move `apify_client._errors` to `apify_client.errors`


fnesveda:
We have some error subclasses like `ApifyApiError` defined in https://github.com/apify/apify-client-python/blob/master/src/apify_client/_errors.py, with the underscore suggesting it's a private submodule.

We have them [documented in the docs](https://docs.apify.com/api/client/python/reference/class/ApifyApiError), though, suggesting people should use them in their `isinstance` checks etc, which they should be able to, since the thrown errors should be a part of the public API of a module.

We should move them out of the private `_errors` submodule to a public `errors` submodule, to make it clear that these are OK to use by end users.

PR #468, opened 2025-08-13
feat: Expose apify_client.errors module

Mantisus:
- Move apify_client._errors to apify_client.errors

### Issues

- Closes: #158


vdusek (changes requested):
Nice, but please let's move `is_retryable_error` function to `_utils` as this should probably stay private.

vdusek (changes requested):
Also, what about `httpx` here? I believe it should be `impit` and its errors.

Mantisus:
> Also, what about httpx here? I believe it should be impit and its errors.

That's right, but when I was preparing the PR, we hadn't merged the PR with impit yet. Updated :)

vdusek (approved):
LGTM

PR #475, opened 2025-08-14
chore: Update to apify-shared v2.0

vdusek:


Pijukatel (approved):
`maybe_extract_enum_member_value` is defined in sdk as well

PR #480, opened 2025-08-20
feat: Add `unqiue_key` based methods to `RequestQueueClient`

Pijukatel:
### Description

- Add `unqiue_key` based methods to `RequestQueueClient` as an alternative to `id` based methods
- Add function to transform  `unqiue_key` to `id`

### Issues

- Part of: https://github.com/apify/crawlee-python/issues/1358

### DIFF
--- a/src/apify_client/clients/resource_clients/request_queue.py
+++ b/src/apify_client/clients/resource_clients/request_queue.py
@@ -226,7 +226,7 @@ class RequestQueueClient(ResourceClient):
         https://docs.apify.com/api/v2#/reference/request-queues/request/get-request
 
         Args:
-            request_unique_key: unique key of the request to retrieve.
+            request_unique_key: Unique key of the request to retrieve.
 
         Returns:
             The retrieved request, or None, if it did not exist.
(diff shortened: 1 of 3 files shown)


src/apify_client/clients/resource_clients/request_queue.py
```diff
@@ -194,6 +220,19 @@ def get_request(self, request_id: str) -> dict | None:
 
         return None
 
+    def get_request_by_unique_key(self, request_unique_key: str) -> dict | None:
```
vdusek, line 223: I believe it's clear that it is the request's unique key (this applies to all occurrences).

```suggestion
    def get_request_by_unique_key(self, unique_key: str) -> dict | None:
```

vdusek (changes requested):

vdusek (approved):
LGTM

Honza just raised some concerns about whether this should be here or in the SDK, so let's wait for his confirmation as well (both options work for me).

Pijukatel:
Will be handled in sdk instead

PR #513, opened 2025-10-08
feat: add started_before and started_after to run list

danpoletaev:
This PR adds new options to `RunCollectionClient.list()`: `started_before` and `started_after`.

Same PR in JS: https://github.com/apify/apify-client-js/pull/763

### DIFF
--- a/src/apify_client/_http_client.py
+++ b/src/apify_client/_http_client.py
@@ -78,7 +78,7 @@ class _BaseHTTPClient:
             elif isinstance(value, list):
                 parsed_params[key] = ','.join(value)
             elif isinstance(value, datetime):
-                utc_aware_dt = value.replace(tzinfo=timezone.utc)
+                utc_aware_dt = value.astimezone(timezone.utc)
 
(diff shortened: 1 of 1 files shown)


src/apify_client/_http_client.py
```diff
@@ -76,6 +77,13 @@ def _parse_params(params: dict | None) -> dict | None:
             # Our API needs lists passed as comma-separated strings
             elif isinstance(value, list):
                 parsed_params[key] = ','.join(value)
+            elif isinstance(value, datetime):
+                utc_aware_dt = value.replace(tzinfo=timezone.utc)
```
barjin, line 81: This only replaces the timezone without actually computing the difference, right? 

So if I e.g. submit a `datetime.now()` to this method, I would only get runs from two hours in the future (I'm on Central European Summer Time).

image

barjin (commented):
That was fast! :) I'm not a Python dev, but the timezone handling IMO needs a bit more care here.

PR #522, opened 2025-10-21
chore: replace `json.loads()` with the internal impit method `Response.json()`

Mantisus:
Replace `json.loads()` with the internal impit method `Response.json()`.


Mantisus:
I think it's worth waiting for the release of `impit` with Python 3.14 support. Then it can be released jointly with PR #520

PR #559, opened 2025-12-13
chore(ci): Align GitHub workflows across Python projects

vdusek:


.github/workflows/_release_pre.yaml
```diff
@@ -88,6 +43,7 @@ jobs:
       APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
 
   publish_to_pypi:
+    if: "!startsWith(github.event.head_commit.message, 'ci') && !startsWith(github.event.head_commit.message, 'docs')"
```
janbuchar: I'd prefer conditions like this to be in the `on_master.yaml` workflow so that it's easier to understand what happens under which circumstances without digging in the referenced files.

PR #573, opened 2026-01-08
fix: fetch key-value store records as attachments

mvolfik:
See apify/apify-core#25006 - fetching with `attachment=true` now allows getting unmodified HTML, without the script snippet for disabling cookies. There are no other side effects (except that the response will have `Content-Disposition: attachment`, but that is irrelevant to clients other than web browsers)

### DIFF
--- a/src/apify_client/clients/resource_clients/key_value_store.py
+++ b/src/apify_client/clients/resource_clients/key_value_store.py
@@ -126,7 +126,7 @@ class KeyValueStoreClient(ResourceClient):
             response = self.http_client.call(
                 url=self._url(f'records/{key}'),
                 method='GET',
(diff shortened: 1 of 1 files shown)


mvolfik:
@vdusek the CI workflow [broke](https://github.com/apify/apify-client-python/actions/runs/20822286994) on master, because `automatic docs theme update` was pushed just after I merged this PR, and now the 'bump and update changelog' action can't push to master (it would need to force-push, or pull-rebase-push).  
I'm leaving it to you, if something needs to be done, please solve it :)

vdusek:
@mvolfik Yeah, this can happen. The changelog will be updated either with the next commit or automatically as part of the next release, so no action is required. Thanks.

PR #582, opened 2026-01-21
fix: update black to fix ReDoS vulnerability

B4nan:
## Summary
- Addresses Dependabot alert #65 (medium severity)
- Updates `black` to ≥24.3.0 to fix CVE-2024-21503 (Regular Expression Denial of Service)
- Added explicit constraint in dev dependencies

## Test plan
- [x] `uv lock` completes successfully
- [x] black version updated to 26.1.0
- [x] CI tests pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,11 @@ dependencies = [
 
 [dependency-groups]
 dev = [
-    "black>=24.3.0",  # Security fix for CVE-2024-21503 (ReDoS vulnerability)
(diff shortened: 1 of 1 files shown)


B4nan:
Security issues in dev dependencies are also valid, because we use them in our CI where we have various secrets that could leak. But you are right that those issues are less important and often invalid for us. Still, if we can resolve them with a dependency bump, we should do so.

What's the alternative here? I guess waiting for pydoc to update their dependencies? We can surely wait a bit with this, I was mostly testing how claude handles this kind of task, since it will get worse once we include more repositories in those checks.

PR #621, opened 2026-02-17
fix: correct boolean parsing of pagination desc header in list_items

vdusek:
## Summary
- `bool(response.headers['x-apify-pagination-desc'])` always returns `True` for any non-empty string, including `"false"`
- The `desc` field in `ListPage` from `list_items` was always `True` regardless of the actual API response
- Fixed by using explicit string comparison: `.lower() == 'true'` (both sync and async variants)

## Test plan
- [ ] Verify existing unit tests pass
- [ ] Verify `list_items(desc=False)` returns `ListPage` with `desc=False`

### DIFF
--- a/tests/unit/test_dataset_list_items.py
+++ b/tests/unit/test_dataset_list_items.py
@@ -1,0 +1,89 @@
+from __future__ import annotations
+
+import json
+from typing import TYPE_CHECKING
+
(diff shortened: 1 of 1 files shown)


Pijukatel (approved):
This fix would be nice with a test

PR #625, opened 2026-02-18
ci: Trigger doc release on stable release

vdusek:
## Summary
- The stable release pipeline was missing a doc release step, so the changelog on the website was not updated after a stable release
- Added a `doc_release` job to `manual_release_stable.yaml` (same pattern as `doc_release_post_publish` in `on_master.yaml`)
- The job runs after `changelog_update` and `pypi_publish`, using the changelog commit ref to include the updated changelog
- Includes explicit permissions (`contents: write`, `pages: write`, `id-token: write`) since the workflow has top-level `permissions: contents: read`

## Test plan
- [ ] Verify the workflow YAML is valid
- [ ] Trigger a stable release and confirm doc release runs


vdusek:
@janbuchar Could you please confirm that this is the correct approach?

Issue #421, opened 2025-06-16
Handle webhook inputs consistently and in one place


Pijukatel:
Unify Webhook input sanitation and handling. Maybe with Pydantic?

**Context:**
Webhook inputs to the clients are now handled in different ways based on the client used. One path is using [encode_webhook_list_to_base64](https://github.com/apify/apify-client-python/blob/66c5bf8dc6355cde8851e6a227d159a64ac66891/src/apify_client/_utils.py#L121), for example in [ActorClient.start](https://github.com/apify/apify-client-python/blob/66c5bf8dc6355cde8851e6a227d159a64ac66891/src/apify_client/clients/resource_clients/actor.py#L215).

The other one is using [get_webhook_representation](https://github.com/apify/apify-client-python/blob/master/src/apify_client/clients/resource_clients/webhook.py#L24), for example in [WebhookClient.update](https://github.com/apify/apify-client-python/blob/master/src/apify_client/clients/resource_clients/webhook.py#L63).

This creates situation where same responsibility is handled in two different places and it can happen than one of the paths has issues that are already fixed in the other one. For example: https://github.com/apify/apify-client-python/pull/419

PR #663, opened 2026-03-06
feat: Accept Pydantic models as alternatives to dicts in resource client methods

vdusek:
## Changes summary

- Accept `dict | PydanticModel` union types in resource client methods that previously only accepted dict, enabling type-safe usage alongside plain dict input.
- Dicts are validated through Pydantic internally, catching schema errors earlier than the API would.
- Replace the dict-based `encode_webhook_list_to_base64` utility with a structured `WebhookRepresentationList` model that handles validation and base64 encoding.
- Add `RequestInput` and `RequestDeleteInput` models in `_types.py` for RQ operations, matching the actual API schema requirements.

## Issues

Closes: #421 

## Motivation

Improve typing on the input side of resource client methods. We don't want to accept only Pydantic models, which would require users to import additional classes. Instead, methods now accept both dict and the corresponding Pydantic model. This way, users don't have to import anything extra, but can optionally use the model types for type-safe usage and to discover the expected dict shape.

## Next steps
  
Once we fully update the OpenAPI spec (https://github.com/apify/apify-docs/issues/2286), we can remove the manually created models in `_types.py` and replace them with the generated ones.

## Test plan

- [x] CI passes


janbuchar (approved):

janbuchar:
Also, maybe we could replace the `dict` args with generated `TypedDict` types for another layer of type safety.

Pijukatel (approved):

Issue #675, opened 2026-03-18
SubNavbar links don't preserve selected docs version


vdusek:
When on a versioned page (e.g. `reference/1.12/...`), clicking "Docs" or "Changelog" in the subNavbar navigates to the latest stable version instead of staying on the selected version.

**Steps to reproduce:**
1. Go to any API reference page
2. Switch to version 1.12 using the version dropdown
3. Click "Docs" in the subNavbar

**Expected:** Navigate to `docs/1.12/...`
**Actual:** Navigates to `docs/...` (latest stable)

**Root cause:** SubNavbar items in `docusaurus.config.js` have static `to` paths (`docs`, `/reference`, `docs/changelog`) with no version awareness. The fix requires overriding the `Navbar/Content` theme component to detect the current version from the URL and inject the version slug into subNavbar item paths.

This applies to JS repos (both client & SDK) as well.

PR #676, opened 2026-03-18
docs: add doc version for v1.12 and v2.5 fixes

vdusek:
## Summary

- Add versioned docs for v1.12 (latest 1.x) from git tag `v1.12.2`.
- Fix Python version requirement in v2.5 docs (3.11+ → 3.10+).
- Override `DocsVersionDropdownNavbarItem` so the version dropdown navigates to `reference/<version>` instead of `docs/<version>` when on API reference pages.
  - Same as in the JS version.

## Known limitation

- SubNavbar links (Docs/Reference/Changelog) don't preserve the selected version — tracked in #675.
  - This applies to JS repos (both client & SDK) as well.

## Test plan

- [x] Tested locally
- [x] CI passes


barjin (approved):
Since the swizzled component is based on https://github.com/apify/apify-client-js/blob/master/website/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js, I don't have any other points.

As discussed through Slack, let's extract this to a common component in `docs-theme` eventually.

Issue #618, opened 2026-02-17
Automate PR creation when OpenAPI spec is updated


vdusek:
### Context

As part of the effort to introduce fully typed & generated API clients ([apify/apify-core#25186](https://github.com/apify/apify-core/issues/25186)), the Python client will generate Pydantic models and type annotations from the [OpenAPI specification](https://docs.apify.com/api/openapi.json) (see #481).

Once the generation pipeline is in place, we need a mechanism to keep the generated code in sync with the OpenAPI spec automatically, so that changes to the API are promptly reflected in the client without manual intervention.

### Goal

Set up a CI/CD automation (e.g. GitHub Actions workflow) that:

1. **Detects changes** to the OpenAPI specification (e.g. via a scheduled check, a webhook/dispatch from `apify-docs`, or by polling `https://docs.apify.com/api/openapi.json`).
2. **Re-runs the code generation** pipeline to produce updated models/types.
3. **Opens a PR** automatically in `apify-client-python` with the regenerated code, so it can be reviewed and merged.

### Considerations

- The trigger mechanism needs to be decided — options include:
  - A **`repository_dispatch`** event sent from the `apify-docs` repo (or wherever the OpenAPI spec is maintained) whenever the spec changes.
  - A **scheduled workflow** (e.g. daily cron) that fetches the latest spec, compares it against the currently used version, and opens a PR if there are differences.
  - A **webhook** from the API deployment pipeline.
- The PR should include a clear diff of generated changes and ideally reference which OpenAPI spec version/commit triggered the update.
- Consider running linting, type checking, and tests as part of the PR to catch breaking changes early.
- Similar automation should likely be set up for `apify-client-js` as well ([apify/apify-client-js#823](https://github.com/apify/apify-client-js/issues/823)).

### Related

- Parent epic: [apify/apify-core#25186](https://github.com/apify/apify-core/issues/25186)
- Typed Python client: #481
- OpenAPI sync mechanism: [apify/apify-core#22169](https://github.com/apify/apify-core/issues/22169)

PR #694, opened 2026-03-31
ci: add workflow to regenerate models from OpenAPI spec

vdusek:
## Summary

- Add `workflow_dispatch` workflow that regenerates Pydantic models from an `apify-docs` PR's OpenAPI spec
- Builds the spec locally, runs `datamodel-codegen`, and opens a PR if models changed
- Comments back on the source `apify-docs` PR with a link

## Issues

- Closes: https://github.com/apify/apify-client-python/issues/618

## Context

- https://github.com/apify/apify-docs/pull/2381
- https://github.com/apify/apify-client-python/pull/694

### DIFF
--- a/.github/workflows/manual_regenerate_models.yaml
+++ b/.github/workflows/manual_regenerate_models.yaml
@@ -168,11 +168,6 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
         run: |
-          if [[ -z "$DOCS_PR_NUMBER" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/regenerate_models.yaml
```diff
@@ -0,0 +1,168 @@
...
+          - apify-docs PR: ${DOCS_PR_URL}
+          EOF
+            )
+
+            PR_URL=$(gh pr create \
+              --title "$TITLE" \
+              --body "$BODY" \
+              --head "$BRANCH" \
+              --base master)
+            echo "Created PR: $PR_URL"
```
janbuchar: These steps should use the commonly used reusable actions (`add-and-commit` and so on)

PR #705, opened 2026-04-07
ci: Use poe tasks for model generation in CI workflow

vdusek:
## Summary

- Add `generate-models-from-file` poe task that accepts a local file path, working around `datamodel-codegen` ignoring `--input` when `url` is set in `pyproject.toml`
- Update CI workflow to call poe tasks instead of `datamodel-codegen` directly, ensuring the downloaded artifact from apify-docs PRs is actually used


janbuchar:
> working around `datamodel-codegen` ignoring `--input` when `url` is set in `pyproject.toml`

This might be worth reporting upstream, right?

janbuchar (approved):

vdusek:
> This might be worth reporting upstream, right?

Yes, I'll do it.

PR #727, opened 2026-04-15
ci: Sort generated model classes alphabetically with topological ordering

vdusek:
## Summary

- Add alphabetical class sorting (with topological ordering to respect inheritance) to `postprocess_generated_models.py`, so that regeneration from a reordered OpenAPI spec produces minimal diffs.
- Make `add_docs_group_decorators` idempotent so the script can be safely re-run.
- Apply the new sorting to `_models.py`.

## Note

- This is a workaround until [datamodel-code-generator#3087](https://github.com/koxudaxi/datamodel-code-generator/issues/3087) is implemented (if ever).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -176,6 +176,7 @@ max-branches = 18
 addopts = "-r a --verbose"
 asyncio_default_fixture_loop_scope = "function"
 asyncio_mode = "auto"
+pythonpath = ["."]
 timeout = 1800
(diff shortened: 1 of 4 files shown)


Pijukatel (commented):
I would suggest adding a few test cases for the sorting. This functionality should not be hard to test.

Pijukatel (approved):

PR #729, opened 2026-04-15
test: guard integration test resource names against 63-char API limit

vdusek:
## Summary

- Adds a 63-character API name limit guard to `get_random_resource_name`, following the approach from `apify-sdk-python`
- Fixes flaky `test_schedule_create_and_get[async]` caused by name collisions with stale schedules from cancelled CI runs

## Failed CI runs

- https://github.com/apify/apify-client-python/actions/runs/24454251436/job/71450645691

### DIFF
--- a/tests/integration/_utils.py
+++ b/tests/integration/_utils.py
@@ -64,9 +64,21 @@ def get_random_string(length: int = 10) -> str:
     return ''.join(secrets.choice(string.ascii_letters) for _ in range(length))
 
 
-def get_random_resource_name(resource: str) -> str:
-    """Generate a random resource name for test resources."""
-    return f'python-client-test-{resource}-{get_random_string(16)}'
+def get_random_resource_name(label: str) -> str:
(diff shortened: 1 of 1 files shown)


tests/integration/_utils.py
```diff
@@ -66,7 +66,7 @@ def get_random_string(length: int = 10) -> str:
 
 def get_random_resource_name(resource: str) -> str:
     """Generate a random resource name for test resources."""
-    return f'python-client-test-{resource}-{get_random_string(5)}'
+    return f'python-client-test-{resource}-{get_random_string(16)}'
```
Pijukatel: I would suggest this approach https://github.com/apify/apify-sdk-python/blob/master/tests/integration/_utils.py#L55, which also ensures that the API limit of 63 characters is not accidentally exceeded

vdusek: makes sense, please recheck

Pijukatel (approved):

Issue #423, opened 2025-06-17
Introduce separate Exception subclasses for each API error type


janbuchar:
- https://apify.slack.com/archives/CD0SF6KD4/p1750081551997999
- This might be solved by code generation

PR #737, opened 2026-04-17
feat: add ApifyApiError subclasses grouped by HTTP status

vdusek:
## Summary

Addresses #423. Adds a small set of `ApifyApiError` subclasses so callers can narrow `except` clauses to specific failure modes. Dispatch is driven by HTTP status — a stable contract — rather than the per-endpoint `error.type` string. `type`, `message`, and `data` remain exposed as metadata.

All subclasses inherit from `ApifyApiError`, so existing `except ApifyApiError` handlers keep matching.

| Class | Status |
|---|---|
| `InvalidRequestError` | 400 |
| `UnauthorizedError` | 401 |
| `ForbiddenError` | 403 |
| `NotFoundError` | 404 |
| `ConflictError` | 409 |
| `RateLimitError` | 429 |
| `ServerError` | 5xx |

Unmapped statuses and unparsable bodies fall back to the base `ApifyApiError`.

## Migration note

Constructing `ApifyApiError(response, ...)` directly now returns an instance of the matching subclass. `except ApifyApiError` is unaffected; tests asserting `type(exc) is ApifyApiError` on a mapped status would need updates. `catch_not_found_or_throw` now uses `isinstance(exc, NotFoundError)` instead of a hardcoded `error.type` allowlist.


vdusek:
Let me guys know whether you think this is worth it. Currently, I'm slightly more inclined not to do it and keep the status quo. Since the API type errors are just general strings, they can often change.

janbuchar:
## Arguments in favor

1. Better DX - easier to catch individual types of exception, while you can still just catch `ApifyApiError` if you don't care much
2. Better DX - the exception name has a non-generic name, and that's the first thing you look at when you see an exception traceback in logs

## Arguments against

1. It' a lot of generated code
2. We need to maintain the generator
3. If an error type disappears from a response object, it might get deleted from the code base, which is a breaking change (but it's also a breaking change in the API, so I'm not sure about the validity of this concern)

Pijukatel:
I would add one convenience argument against this change, which would only be used as a tie breaker:

Adding this feature is non-breaking. Removing it is breaking. 

So I would wait until we are more certain about it, as we can easily add this in any minor release.

src/apify_client/errors.py
```diff
@@ -35,6 +47,18 @@ class ApifyApiError(ApifyClientError):
...
+        """Dispatch to the subclass matching the response's HTTP status code, if any."""
+        target_cls: type[ApifyApiError] = cls
+        if cls is ApifyApiError:
+            status = response.status_code
+            mapped = _STATUS_TO_CLASS.get(status)
+            if mapped is None and status >= HTTPStatus.INTERNAL_SERVER_ERROR:
+                mapped = ServerError
+            if mapped is not None:
+                target_cls = mapped
+        return super().__new__(target_cls)
```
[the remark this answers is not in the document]
vdusek: addressed

src/apify_client/errors.py
```diff
@@ -43,27 +67,17 @@ def __init__(self, response: HttpResponse, attempt: int, method: str = 'GET') ->
...
-                if 'data' in response_data['error']:
-                    self.data = response_data['error']['data']
-
-        except ValueError:
-            pass
+        if payload is not None:
+            self.message = payload['message']
+            self.type = payload['type']
+            if 'data' in payload:
+                self.data = payload['data']
```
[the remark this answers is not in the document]
vdusek: addressed

src/apify_client/_utils.py
```diff
@@ -63,9 +62,7 @@ def catch_not_found_or_throw(exc: ApifyApiError) -> None:
     Raises:
         ApifyApiError: If the error is not a 404 Not Found error.
     """
-    is_not_found_status = exc.status_code == HTTPStatus.NOT_FOUND
-    is_not_found_type = exc.type in ['record-not-found', 'record-or-token-not-found']
-    if not (is_not_found_status and is_not_found_type):
+    if not isinstance(exc, NotFoundError):
```
Pijukatel, line 65: I am not sure about this (but the previous variant had the same issues.)

The problem of swallowing `404`s is that you lose the information about what was not found. For example

```
dataset_1 = client.run(run_id="RUN DOES NOT EXIST").dataset().get() # Run does not exist
dataset_2 = client.run(run_id="RUN EXISTS").dataset().get() # Dataset does not exist
```
In both cases it returns `None` and user has no idea which of the cases happened.

Pijukatel: We should probably just swallow 404 related to the last resource in the chain and throw all the others?

vdusek: Thanks for bringing this up.

> We should probably just swallow 404 related to the last resource in the chain and throw all the others?

I implemented it this way. It's certainly better, but I'm not sure in terms of consistency - now for the get we for some cases return None and for some other raise exception, hmm?

Pijukatel: Ok, lets discuss this in the team https://apify.slack.com/archives/C02JQSN79V4/p1776845186944399

Pijukatel (approved):
I am fine with both 404 variants.

src/apify_client/_utils.py
vdusek: Let's remove the last commit, merge it as it is, and do this in a dedicated one if there is a consensus.

PR #736, opened 2026-04-17
refactor!: Use snake case for `actors().list(sort_by=...)`

vdusek:
## Summary

- Changes the `sort_by` parameter of `ActorCollectionClient(Async).list()` from the camelCase API values (`'createdAt'`, `'stats.lastRunStartedAt'`) to pythonic snake_case values (`'created_at'`, `'last_run_started_at'`).
- Adds a small module-level `_SORT_BY_TO_API` mapping to translate to the API's expected form before dispatching the request.
- Updates the integration test accordingly.

Closes #700.

Issue #576, opened 2026-01-09
Prefer using literal types over enums


vdusek:
- We should prefer `typing.Literal` types over `enum.Enum`.

### Motivation

- Better developer experience - Literals can be passed directly as values, without requiring additional imports.
- Type safety is preserved - When used with a type checker, Literal types provide the same level of safety as enums.
- Fewer footguns - The usage of enums can easily lead to bugs like this one: https://github.com/apify/apify-client-python/pull/575.
- Consistency across our tooling - We already prefer literals in Crawlee and the SDK.

### Scope

- This primarily applies to constants currently defined as enums, especially those in: https://github.com/apify/apify-shared-python/blob/master/src/apify_shared/consts.py.

### Compatibility

- This is a breaking change, so it should be tracked and implemented as part of apify-python-client v3.

PR #738, opened 2026-04-17
feat: generate TypedDict types for input-side models

### LINKED ISSUE #666: Generate TypedDict types for input-side models
## Context

In PR #663 (Accept Pydantic models as alternatives to dicts in resource client methods), we introduced union types like `dict | WebhookCreate` so that users can pass Pydantic model instances instead of plain dicts. However, the `dict` part of the union is still untyped — users passing dicts get no type safety or autocompletion.

As discussed in [this comment](https://github.com/apify/apify-client-python/pull/663#issuecomment-4037713076), we could replace the generic `dict` with generated `TypedDict` types for an additional layer of type safety. For example, instead of `dict | WebhookCreate`, the signature could be `WebhookCreateDict | WebhookCreate`, where `WebhookCreateDict` is a `TypedDict` with the same fields.

## Proposal

- Investigate generating `TypedDict` equivalents from the Pydantic models (or directly from the OpenAPI spec) for models that are used as input parameters.
- Integrate the generation into the existing model generation pipeline (`uv run poe generate-models`).
- Replace `dict` with the corresponding `TypedDict` in resource client method signatures.

vdusek:
## Summary

Adds `TypedDict` counterparts for input-side models so users passing plain dicts to resource-client methods get full type-checker support without importing Pydantic models.

- Generate TypedDicts alongside Pydantic models via a second `datamodel-code-generator` pass (`--output-model-type typing.TypedDict`) in `poe generate-models`.
- Split generated content into `_models_generated.py` / `_typeddicts_generated.py`. Hand-written `_models.py` / `_typeddicts.py` now only hold shapes not exposed by the OpenAPI spec (e.g. `RequestInput`, `RequestInputDict`).
- Extend `scripts/postprocess_generated_models.py` to trim the TypedDict file to input-relevant classes (plus transitive deps), rename them with a `Dict` suffix, and add `@docs_group('Typed dicts')`.
- Widen resource-client input parameters (`actor`, `task`, `task_collection`, `request_queue`) to accept `TypedDict | PydanticModel` unions.
- Update `.rules.md` and `manual_regenerate_models.yaml` to reflect the new layout.

Closes #666.


vdusek:
@Pijukatel @janbuchar Let me guys know what you think, this change also has some drawbacks:

1. The postprocess script had to be extended significantly.
2. Some models and TypedDicts still need to be maintained manually, so I split them into two modules (generated vs. hand-written). I am not sure about this structure.

tests/integration/test_request_queue.py
```diff
@@ -75,7 +77,12 @@ async def test_request_queue_lock(client: ApifyClient | ApifyClientAsync, *, is_
         # Add requests and check if correct number of requests was locked
         for i in range(15):
             await maybe_await(
-                rq.add_request({'url': f'http://test-lock.com/{i}', 'uniqueKey': f'http://test-lock.com/{i}'})
+                rq.add_request(
```
Pijukatel: Why do we need this cast? Isn't this already compliant with the `RequestInputDict`?

I tried locally, and it was not complaining when the cast was removed. (same for other casts as well)

docs/02_concepts/code/03_nested_sync.py
```diff
@@ -1,5 +1,5 @@
 from apify_client import ApifyClient
-from apify_client._models import ActorJobStatus
```
janbuchar, line 2: Not really a problem with this PR, but are we promoting importing from `_models` (internal module) in the docs? 🤔

src/apify_client/_typeddicts_generated.py
```diff
@@ -0,0 +1,102 @@
...
+    unique_key: NotRequired[str]
+    """
+    A unique key used for request de-duplication. Requests with the same unique key are considered identical.
+    """
+    url: NotRequired[str]
+    """
+    The URL of the request.
+    """
+    method: NotRequired[Literal['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']]
+    retry_count: NotRequired[int]
```
janbuchar, line 21: I think there is a case for accepting camelCase here instead - it is exactly what the API accepts. On the other hand, snake_case fits in better with the rest of the Python code and works better for the "I just don't want to look up what model I need to import" use case for TypedDicts. Did you consider this @vdusek?

src/apify_client/_types.py
```diff
@@ -18,129 +12,5 @@
 
 JsonSerializable = str | int | float | bool | None | dict[str, Any] | list[Any]
 """Type for representing json-serializable values. It's close enough to the real thing supported by json.parse.
-It was suggested in a discussion with (and approved by) Guido van Rossum, so I'd consider it correct enough.
+It was suggested in a discussion with (and approved by) Guido van Rossum, so let's go with it.
```
janbuchar: Okay? 😁

vdusek: This is some remnant from the past, maybe just remove it completely :smile:

docs/02_concepts/code/03_nested_sync.py
vdusek: Yeah, this is probably because of enums (see https://github.com/apify/apify-client-python/issues/576), but `Literal` types aren't very well supported by `datamodel-code-gen`, so it might require some additional post-processing. Let's see.

Also, the `TypedDict`s should be public as well. I'll do a final revision before the v3 release.

vdusek: issue - https://github.com/apify/apify-client-python/issues/744

tests/integration/test_request_queue.py
vdusek: thanks

src/apify_client/_types.py
vdusek: Okay, I updated the type as well -> https://github.com/apify/apify-client-python/pull/745, it is not true for a long time that recursive types are not supported by mypy (or other type checkers).

src/apify_client/_typeddicts_generated.py
vdusek: I think we should keep the public interface Python-idiomatic and stick with `snake_case`. I recently made some updates in this direction, moving the interface to `snake_case`, see [https://github.com/apify/apify-client-python/pull/736](https://github.com/apify/apify-client-python/pull/736).

I wouldn't have a bug problem in leaving it just as the API accepts it, but we would have to do it in all places to be consistent.

scripts/postprocess_generated_models.py
```diff
@@ -47,46 +68,236 @@ def deduplicate_error_type_enum(content: str) -> str:
...
+    refs: set[str] = set()
+    for child in ast.walk(node):
+        if isinstance(child, ast.Name):
+            refs.add(child.id)
+    return refs - exclude
+
+
+def _compute_transitive_closure(deps: dict[str, set[str]], seeds: set[str]) -> set[str]:
+    """Return every symbol transitively reachable from any seed."""
+    closure: set[str] = set()
```
Pijukatel: Why is this set called like this? Maybe pick a better name that would indicate what it is supposed to contain

scripts/postprocess_generated_models.py
```diff
@@ -1,31 +1,52 @@
...
 # Map of camelCase discriminator values to their snake_case equivalents.
 # Add new entries here as needed when the OpenAPI spec introduces new discriminators.
 DISCRIMINATOR_FIXES: dict[str, str] = {
     'pricingModel': 'pricing_model',
 }
 
+# Seed models for the TypedDict pruning. Every TypedDict in `_typeddicts.py` that is not
+# transitively reachable from this set is removed. Keep in sync with the `dict | <Model>` unions
+# on resource-client method signatures.
+TYPEDDICT_SEEDS: frozenset[str] = frozenset(
```
Pijukatel, line 42: The naming is more about the pruning algorithm than about the top-level intent. Why are these three selected?

Maybe some name like:
`REQUIRED_DICTS` or `USED_DICTS` ...

and some hint in the comment about why just these 3 are selected

src/apify_client/_typeddicts_generated.py
janbuchar: Since we shove everything into Pydantic, we should also be perfectly capable of supporting both, no? Like, camelCase and snake_case version of every model where this is applicable.

scripts/postprocess_generated_models.py
```diff
@@ -47,46 +68,236 @@ def deduplicate_error_type_enum(content: str) -> str:
...
+def _collect_name_references(node: ast.AST, exclude: set[str]) -> set[str]:
+    """Collect every `ast.Name` identifier under `node`, minus names in `exclude`."""
+    refs: set[str] = set()
+    for child in ast.walk(node):
+        if isinstance(child, ast.Name):
+            refs.add(child.id)
+    return refs - exclude
+
+
+def _compute_transitive_closure(deps: dict[str, set[str]], seeds: set[str]) -> set[str]:
```
Pijukatel, line 194: Usage of `closure` in naming can be misleading, as it has its own meaning in Python, different from set theory, and while reading Python code, the reader will probably have the Python meaning in mind first.

Maybe better name based on what is the purpose of it: 
`_compute_reachable_symbols` or `_compute_transitive_dependencies` or ...

and same for `closure: set[str] = set()` -> `reachable_symbols` or `transitive_dependencies` or...

tests/unit/test_utils.py
```diff
@@ -54,7 +58,11 @@ def test_webhook_representation_list_to_base64() -> None:
 
 def test_webhook_representation_list_from_dicts() -> None:
     """Test that from_webhooks accepts plain dicts with the minimal ad-hoc webhook shape."""
-    result = WebhookRepresentationList.from_webhooks(
+    # The runtime only needs the keys consumed by WebhookRepresentation (event_types, request_url,
```
Pijukatel, line 61: If it does not require so much, is the typing correct?

src/apify_client/_typeddicts_generated.py
vdusek: I don't think so, TypedDicts don't have anything like `Field(alias=...)`.

AI summary of what we do with the models:

> The generated Pydantic models use `model_config = ConfigDict(populate_by_name=True)` and each field has a camelCase `Field(alias=...)`. `populate_by_name=True` tells Pydantic "when constructing or validating, accept either the Python field name (snake_case) or the declared alias (camelCase)." So something like `Run.model_validate({'defaultDatasetId':  'x'})` and `Run.model_validate({'default_dataset_id': 'x'})` both work today. Serialization still emits camelCase (`via by_alias=True`) so the wire format matches the API.

But `TypedDicts` don't have the equivalent of Pydantic's `Field(alias=...)` and there's no built-in way to declare "this key may be either unique_key or uniqueKey" in a single TypedDict. We would have to generate a 2nd TypedDict for every model with camelCase keys, and then combined it via union in the input. IMO that would be horrible.

scripts/postprocess_generated_models.py
vdusek: updated

src/apify_client/_typeddicts_generated.py
janbuchar: > We would have to generate a 2nd TypedDict for every model with camelCase keys, and then combined it via union in the input.

Yeah, that's what I had in mind.

> IMO that would be horrible.

Care to elaborate? :grin: I agree that it would increase the amount of generated code by quite a bit, but it could also be more approachable for users (even from a BC point of view).

vdusek: Hmm, okay, so a bit more overhead for fewer breaking changes. OK, it might be reasonable trade-off. 

So something like this?

```python
request: Request | RequestDict | RequestDictCamelCase
```

And do we wanna deprecate `RequestDictCamelCase` with a warning, and remove it for v4, or keep it there until the end of the world? :slightly_smiling_face:

janbuchar: Not sure if there is a type checker that would pick up the deprecation... But we both know that there's no way that we'll remove this once it's in :grinning:

vdusek: I'm not saying no, but this also overlaps with other type arguments - not just dicts, but literals as well. See https://github.com/apify/apify-client-python/pull/736 for example.

If we want to support both `snake_case` and `camelCase` variants, we should apply it consistently across all input types. I'd suggest not doing it in this PR and instead addressing it as a separate issue - https://github.com/apify/apify-client-python/issues/756.

It's labeled as "consider" and "plan" because I'm not yet sure how literals will be handled, they’re still pending, see https://github.com/apify/apify-client-python/issues/576.

Are you okay with that?

janbuchar: Makes sense, thank you

janbuchar (approved):

tests/unit/test_utils.py
vdusek: Updated, tested against the live API, it requires only `eventTypes` and `requestUrl`. I added `WebhookRepresentationDict` - a minimal TypedDict with only `event_types` and `request_url` required. It mirrors the existing `WebhookRepresentation` model. The `WebhookRepresentationList.from_webhooks` was also adjusted.

Pijukatel (approved):

PR #741, opened 2026-04-18
ci: Fix docs release failing on detached HEAD

vdusek:
## Summary

The `doc_release_post_publish` job in `on_master.yaml` invokes `manual_release_docs.yaml` with a commit SHA as `ref`, so `actions/checkout` leaves the repo in a detached HEAD state. The hand-rolled `Commit the updated package.json and lockfile` step then runs `git push`, which fails with `fatal: You are not currently on a branch.` (see [failing run](https://github.com/apify/apify-client-python/actions/runs/24564853529/job/71822812783)). The push was also invoked unconditionally even when there was nothing to commit.

## Fix

- Replace the hand-rolled git block with `EndBug/add-and-commit@v10` (matching `manual_version_docs.yaml`).
- Set `new_branch: ${{ github.event.repository.default_branch }}` so the action creates/checks out a local default branch before committing. This is required because EndBug does not handle detached HEAD by itself — its default push is `git push origin --set-upstream` (no refspec), which fails the same way.
- Apply the same `new_branch` fix to `manual_version_docs.yaml`, which has the identical latent bug (introduced in #728 but not yet exercised by a stable release).

### DIFF
--- a/.github/workflows/manual_release_docs.yaml
+++ b/.github/workflows/manual_release_docs.yaml
@@ -1,20 +1,17 @@
 name: Release docs
 
 on:
-  # Runs when manually triggered from the GitHub UI.
+  # Runs when manually triggered from the GitHub UI. Always checks out the default branch —
+  # this workflow pushes a theme-update commit back to the default branch, so the checked-out
+  # ref must be its current tip to avoid non-fast-forward pushes.
   workflow_dispatch:
-    inputs:
-      ref:
-        description: Git ref to checkout (branch, tag, or SHA). Defaults to the default branch.
-        required: false
-        type: string
-        default: ""
 
-  # Runs when invoked by another workflow.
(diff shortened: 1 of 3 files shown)


Pijukatel (commented):
I am not so sure what is the desired behavior of using `ref` argument.

- When it is the same as the tip of the master (making the argument useless), then the workflow works fine as the commit will succeed. 
- When it is some random ref that is not on the master,  it will most likely fail pushing to master. Or?

vdusek:
@Pijukatel You're right.  The `ref` argument doesn't make sense for `workflow_dispatch`, but it's needed for `workflow_call` to move HEAD to a specific commit (because of changelog, version docs, and apify theme commits). I've dropped the ref input from `workflow_dispatch` (manual runs now always use the default branch) and kept it as required for `workflow_call`. Thanks.

Pijukatel (approved):

PR #743, opened 2026-04-20
feat: add filter and cursor parameters to list_requests method

mvolfik:
add support for endpoint parameters and new pagination method added in https://github.com/apify/apify-core/issues/12115

There is a new parameter `filter` for the _request queue -> list requests_ endpoint. To properly support pagination, we also added proper cursor-based pagination, with opaque cursor provided by the server.

### DIFF
--- a/src/apify_client/_resource_clients/request_queue.py
+++ b/src/apify_client/_resource_clients/request_queue.py
@@ -5,7 +5,7 @@ import math
 import warnings
 from collections.abc import Iterable
 from queue import Queue
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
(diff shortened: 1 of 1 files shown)


vdusek (approved):
LGTM, just wondering, since we're planning a new major release, shouldn't we mention the `exclusive_start_id` -> `cursor` change in the upgrade guide?

mvolfik:
I mean, this is more of a underlying API change, unrelated to the Client release cycle, and it is completely backwards-compatible, so I'm not sure if it's worth cluttering the upgrade guide. This won't break anything, so people don't need to truly care about it, but the will notice the deprecation warnings, and eventually fix it themselves, no?

vdusek:
alright, let's go

PR #745, opened 2026-04-21
refactor: Improve JsonSerializable type to be recursive

vdusek:
## Summary

Replace the loose `dict[str, Any] | list[Any]` variants of `JsonSerializable` with recursive references so nested JSON values are properly typed instead of falling back to `Any`. Based on the definition discussed in https://github.com/python/typing/issues/182.

### DIFF
--- a/src/apify_client/_types.py
+++ b/src/apify_client/_types.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 from base64 import b64encode
 from datetime import timedelta
-from typing import Annotated, Any, Literal
+from typing import Annotated, Literal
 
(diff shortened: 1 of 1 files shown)


janbuchar (commented):
This is part of the public interface via `UserData`, right? So the change is technically breaking for users of type checkers that cannot handle recursive types. Do we consider it a problem?

vdusek:
See [this comment](https://github.com/python/typing/issues/182#issuecomment-1320974824) from the end of 2022:

> All major type checkers now support recursive type aliases by default, so this should largely work:
> ```
> JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None
> ```

So I wouldn't be afraid of it.

janbuchar (approved):
LGTM then!

Issue #539, opened 2025-11-19
Add better support for pagination


Pijukatel:
Parity with JS client:
https://github.com/apify/apify-client-js/issues/777

vdusek:
During the implementation, please make sure these points are considered, and either addressed or evaluated as false positives.

---

## 2. `iterate_items` early-terminates when filters shrink the page

- **Severity:** normal
- **File:** `src/apify_client/_resource_clients/dataset.py:295-304`

`iterate_items` terminates early whenever `skip_empty`, `skip_hidden`, or `clean=True` causes a server-side page to contain fewer than 1000 items — even when there are many more items left in the dataset. This results in **silent data loss** for any caller using the documented filter parameters.

```python
current_page_item_count = len(current_items_page.items)
read_items += current_page_item_count

if current_page_item_count < cache_size:
    should_finish = True
```

This is only correct when every item in the scanned `[offset, offset+limit)` window is returned verbatim. The Apify API applies `skip_empty`, `skip_hidden`, and `clean` as **post-filters**, so a short response does **not** imply the dataset is exhausted.

### Step-by-step proof

Dataset with 5000 items, ~half empty, caller invokes `dataset.iterate_items(skip_empty=True)`:

1. First: `list_items(offset=0, limit=1000, skip_empty=True)` → server scans 0..999, filters out ~500 empties, returns ~500.
2. `current_page_item_count = 500`; `read_items = 500`.
3. `500 < 1000` → `should_finish = True`.
4. Loop exits. ~2000 remaining non-empty items are **silently dropped**.

### Secondary: offset drift

Line 285 advances the window using `read_items` (items returned) rather than items scanned: `offset=offset + read_items`. Under any filter, `read_items < items scanned`, so the next page re-reads items already scanned.

### Fix

Track the scanned window explicitly, or terminate only on an empty response (`current_page_item_count == 0`).

---

## 3. `iterate_items` does not forward `flatten` and `view`

- **Severity:** normal
- **File:** `src/apify_client/_resource_clients/dataset.py:284-296`

[...]

PR #747, opened 2026-04-21
fix: Prevent `iterate_items` from terminating early under server-side filters

vdusek:
## Summary

- `DatasetClient.iterate_items` (both sync and async) terminated early whenever `skip_empty`, `skip_hidden`, or `clean=True` caused a server-side page to contain fewer than 1000 items, silently dropping the remaining dataset items.
- The API applies those filters **after** scanning the `[offset, offset+limit)` window, so a short page does not imply the dataset is exhausted. The previous `current_page_item_count < cache_size` shortcut was incorrect under any server-side filter.
- The request offset was also advanced by the number of items **returned** rather than **scanned**, causing subsequent requests to re-scan overlapping windows whenever a filter was active.

## Fix

- Track the scanned window explicitly via a new `scanned_items` counter and use it to compute the next request offset.
- Advance `scanned_items` by `effective_limit` (the size of the requested window), not by the count of returned items.
- Terminate only when the server returns zero items (or the caller-supplied `limit` has been reached).

## Tests

- Added regression tests (sync and async) that simulate an API post-filtering half of every scanned 1000-item window across a 2500-item dataset. They fail on `master` (loop exits after the first ~500 items) and pass on this branch (all ~1250 non-empty items are yielded).

### DIFF
--- a/src/apify_client/_resource_clients/dataset.py
+++ b/src/apify_client/_resource_clients/dataset.py
@@ -268,13 +268,16 @@ class DatasetClient(ResourceClient):
         """
         cache_size = 1000
 
-        should_finish = False
         read_items = 0
+        # Server-side filters (skip_empty, skip_hidden, clean) drop items after the [offset, offset+limit)
+        # window is scanned, so a short page does not imply the dataset is exhausted. Track the scanned
+        # window explicitly and stop only when the server returns no items at all.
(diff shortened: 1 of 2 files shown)


vdusek:
Closing as this part of codebase will be refactored during https://github.com/apify/apify-client-python/issues/539

PR #751, opened 2026-04-22
fix: Preserve count=0 in RunClient's charge

vdusek:
## Summary

`RunClient.charge` (sync + async) built the request body with `'count': count or 1`. Because `or` treats `0` as falsy, a caller passing `count=0` — e.g. `count=len(batch)` for an empty batch, or a defensive no-op — was silently sent as `count=1`. On a pay-per-event billing endpoint that's a real overcharge, not a cosmetic bug.

## Fix

Changed the signature from `count: int | None = None` to `count: int = 1` and pass `count` through to the request body unchanged. `0` now reaches the server as `0`; the default `1` is expressed in the signature instead of via a fallback.

Docstring updated to reflect the new contract.

## Tests

`tests/unit/test_run_charge.py` parametrizes the sync and async clients over `count` values `0`, `1`, `5` and asserts the value lands in the request body verbatim. The `0` case fails on `master` and passes here.

### DIFF
--- a/src/apify_client/_resource_clients/run.py
+++ b/src/apify_client/_resource_clients/run.py
@@ -385,7 +385,8 @@ class RunClient(ResourceClient):
 
         Args:
             event_name: The name of the event to charge for.
-            count: The number of events to charge. Defaults to 1 if not provided.
+            count: The number of events to charge. Defaults to 1 when `None`; other values,
+                including 0, are sent to the server as-is.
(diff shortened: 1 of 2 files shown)


src/apify_client/_resource_clients/run.py
Pijukatel: Why don't we fix it here?

`count: int = 1,`

Pijukatel (changes requested):

Pijukatel (approved):

PR #759, opened 2026-04-23
feat!: generate Literal type aliases instead of StrEnum classes

### LINKED ISSUE #576: Prefer using literal types over enums

vdusek:
## Issue

- Closes #576

## Description

Replaces every generated `class X(StrEnum)` with a `X = Literal[...]` alias. Users pass plain strings (`status='RUNNING'`) instead of enum members, the aliases are reusable across method signatures, and behavior matches Crawlee and the Apify SDK — avoiding mixing-enum-with-string footguns like #575.

## Why post-processing

The natural fix — `--enum-field-as-literal=all` + `--use-type-alias` on `datamodel-code-generator` — only handles enums referenced from operation parameters; the rest get inlined at every reference site with no stable name to import. Filed upstream as [koxudaxi/datamodel-code-generator#3104](https://github.com/koxudaxi/datamodel-code-generator/issues/3104). Until that lands, `scripts/postprocess_generated_models.py` does the rewrite ourselves; once fixed, we can drop the post-processing.

## What the post-process does

- **`convert_enums_to_literals`** — rewrites every top-level `class X(StrEnum)` to `X = Literal[...]`, preserving value order and class docstring.
- **`deduplicate_error_type_enum`** — removes the duplicate inlined `class Type(StrEnum)` datamodel-codegen emits alongside the named `ErrorType` (same upstream issue), and rewires `Type` annotations to `ErrorType`. AST-based.
- **`split_literals_to_file`** — moves the 11 alias blocks into `_literals_generated.py` so consumers don't pull in every Pydantic model.
- **`snake_case_camelcase_literal_values`** — converts camelCase string values (`StorageOwnership`'s `'ownedByMe'`/`'sharedWithMe'`) to snake_case and emits `_<NAME>_WIRE_VALUES` so resource clients can convert back to the wire format.

The hand-maintained `_types.py` is renamed to `_literals.py` for symmetry.

## Consumers

- All internal code imports the aliases via `TYPE_CHECKING` where possible.
- `dataset_collection`, `key_value_store_collection`, `request_queue_collection` translate `ownership` back to wire format via `_STORAGE_OWNERSHIP_WIRE_VALUES`.
- `_TERMINAL_STATUSES` in `_resource_client.py` is derived from `TerminalActorJobStatus` via `typing.get_args()` — single source of truth.
- Docstrings, integration tests, unit tests, and the two `docs/02_concepts/code/03_nested_*.py` examples no longer reference `.MEMBER` / `.MEMBER.value`.

## Tests

`tests/unit/test_postprocess_generated_models.py` covers each step independently plus a full-pipeline integration test.

## Breaking change

Tracked under v3 on #576. `ActorJobStatus.RUNNING` and similar enum-member access no longer exists — pass the plain string instead.


src/apify_client/_types.py
janbuchar, line 1: This rename is 1) pointless and 2) wrong - the file contains non-literal types too

scripts/postprocess_generated_models.py
```diff
@@ -63,18 +98,220 @@ def fix_discriminators(content: str) -> str:
...
+        isinstance(node, ast.Assign)
+        and len(node.targets) == 1
+        and isinstance(node.targets[0], ast.Name)
+        and isinstance(node.value, ast.Subscript)
+        and isinstance(node.value.value, ast.Name)
+        and node.value.value.id == 'Literal'
+    )
+
+
+def snake_case_camelcase_literal_values(content: str) -> str:
```
janbuchar: Please improve the name so that it's clear what is the input convention and what is on the output

janbuchar (changes requested):

src/apify_client/_types.py
vdusek: Yeah, I wanted to distinguish between generated and non-generated modules. But let's keep literals, models, typeddicts generated, and types manual.

Let me know if you have another (and better) idea.

docs/04_upgrading/upgrading_to_v3.mdx
```diff
@@ -306,3 +306,53 @@ client.actors().list(sort_by='last_run_started_at')
...
+
+```python
+from apify_client._literals_generated import WebhookEventType
+
+events: list[WebhookEventType] = ['ACTOR.RUN.SUCCEEDED', 'ACTOR.RUN.FAILED']
+```
+
+### Snake_case `StorageOwnership` values
+
+`StorageOwnership` previously carried the camelCase wire values (`'ownedByMe'` / `'sharedWithMe'`). It now uses pythonic snake_case values; the client translates them to the camelCase form expected by the API internally.
```
janbuchar: It's honestly weird that we convert `camelCase` string literals to `snake_case`, but we leave stuff like `ACTOR.RUN.SUCCEEDED` or the kebab-case error types live happily ever after. Is that intended?

janbuchar (commented):
In general, I like where this is going, but I feel like we need to align about some inconsistencies.

docs/04_upgrading/upgrading_to_v3.mdx
vdusek: I removed all the string literal transformations, as we discussed on Slack.

tests/integration/test_actor.py
```diff
@@ -60,7 +60,7 @@ async def test_list_actors_pagination(client: ApifyClient | ApifyClientAsync) ->
 
 async def test_list_actors_sorting(client: ApifyClient | ApifyClientAsync) -> None:
```
janbuchar, line 61: For a test that claims to check sorting, it's not very sensitive. If you can change the sort type without the test breaking, then it should probably try harder.

janbuchar (approved):
I left one comment that is not really relevant to this PR - please either dismiss it or make an issue to track it.

tests/integration/test_actor.py
vdusek: https://github.com/apify/apify-client-python/issues/773

PR #769, opened 2026-04-29
feat: Make list methods of clients iterable

Pijukatel:
### Description
- All collection clients list method returns an iterator as well. 
- All async collection clients list method returns an async iterator as well. 
- List of modified clients (same for async clients):
  - ActorCollectionClient
  - BuildCollectionClient
  - RunCollectionClient
  - ScheduleCollectionClient
  - TaskCollectionClient
  - WebhookCollectionClient
  - WebhookDispatchCollectionClient
  - DatasetCollectionClient
  - KeyValueStoreCollectionClient
  - RequestQueueCollectionClient
  - StoreCollectionClient
  - ActorEnvVarCollectionClient (no pagination)
  - ActorVersionCollectionClient (no pagination)
- Additionally, the following storage-related list methods were modified to support iteration as well:
  - DatasetClient.list_items (and marking Dataset.iterate_items as deprecated)
  - KeyValueStoreClient.list_keys (and marking Dataset.iterate_items as deprecated)
  - RequestQueueClient.list_requests

#### Example usage
```
...
# Sync
datasets_client = ApifyClient(token='...').datasets()

# Same as before
list_page = datasets_client.list(...)

# New functionality
individual_items = [item for item in datasets_client.list(...)]

...
# Async
datasets_client = ApifyClientAsync(token='...').datasets()

# Same as before
list_page = await datasets_client.list(...)

# New functionality
individual_items = [item async for item in datasets_client.list(...)]
```

### Issues

- Implements:  #539

### Testing

- Unit tests
- Manual API tests

### Checklist

- [ ] CI passed

### DIFF
--- a/docs/02_concepts/08_pagination.mdx
+++ b/docs/02_concepts/08_pagination.mdx
@@ -12,9 +12,9 @@ import ApiLink from '@site/src/components/ApiLink';
 
 import PaginationAsyncExample from '!!raw-loader!./code/08_pagination_async.py';
 import PaginationSyncExample from '!!raw-loader!./code/08_pagination_sync.py';
-
 import IterateItemsAsyncExample from '!!raw-loader!./code/08_iterate_items_async.py';
 import IterateItemsSyncExample from '!!raw-loader!./code/08_iterate_items_sync.py';
+
 
 Most methods named `list` or `list_something` in the Apify client return a <ApiLink to="class/ListPage">`ListPage`</ApiLink>  object. This object provides a consistent interface for working with paginated data and includes the following properties:
 
@@ -45,7 +45,7 @@ The <ApiLink to="class/ListPage">`ListPage`</ApiLink> interface offers several k
 
 ## Generator-based iteration
 
-For most use cases, `iterate_items()` is the recommended way to process all items in a dataset. It handles pagination automatically using a Python generator, fetching items in batches behind the scenes so you don't need to manage offsets or limits yourself.
+You can also use the `list` methods directly in iteration. It handles pagination automatically, fetching items in batches behind the scenes so you don't need to manage offsets or limits yourself.
 
 <Tabs>
     <TabItem value="AsyncExample" label="Async client" default>
@@ -60,6 +60,4 @@ For most use cases, `iterate_items()` is the recommended way to process all item
     </TabItem>
 </Tabs>
 
-`iterate_items()` accepts the same filtering parameters as `list_items()` (`clean`, `fields`, `omit`, `unwind`, `skip_empty`, `skip_hidden`), so you can combine automatic pagination with data filtering.
-
-Similarly, `KeyValueStoreClient` provides an `iterate_keys()` method for iterating over all keys in a key-value store without manual pagination.
+Similarly, you can iterate over the return value of `KeyValueStoreClient.list_keys()` to go through all keys in a key-value store without manual pagination. The older `iterate_keys()` method is deprecated.
--- a/src/apify_client/_pagination_classes.py
+++ b/src/apify_client/_pagination_classes.py
@@ -1,0 +1,267 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, Generic, TypeVar
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterator, Awaitable, Callable, Generator, Iterator
+
+from apify_client._models_generated import (
+    ActorShort,
+    BuildShort,
+    DatasetListItem,
+    EnvVar,
+    KeyValueStore,
+    KeyValueStoreKey,
+    Request,
+    RequestQueueShort,
+    RunShort,
+    ScheduleShort,
+    StoreListActor,
+    TaskShort,
+    Version,
+    WebhookDispatch,
+    WebhookShort,
+)
+
+T = TypeVar('T')
+
+
+@dataclass
+class IterableOf(Generic[T]):
+    _get_iterator: Callable[[], Iterator[T]] = field(repr=False, compare=False)
+
+    def __iter__(self) -> Iterator[T]:
+        return self._get_iterator()
+
+
+@dataclass
+class AsyncIterableOf(Generic[T]):
+    _get_async_iterator: Callable[[], AsyncIterator[T]] = field(repr=False, compare=False)
+
+    def __aiter__(self) -> AsyncIterator[T]:
+        return self._get_async_iterator()
+
+
+@dataclass
+class PageWithItems(Generic[T]):
+    items: list[T]
+
+
+@dataclass
+class PageWithTotal:
+    total: int
+
+
+@dataclass
+class PageWithLimit:
+    limit: int
+
+
+@dataclass
+class PageWithCount:
+    count: int
+
+
+@dataclass
+class PageWithOffset:
+    offset: int
+
+
+@dataclass
+class PageWithDesc:
+    desc: bool
+
+
+@dataclass
+class PageOfItems(PageWithItems[T], PageWithTotal, PageWithLimit, PageWithCount, PageWithOffset, PageWithDesc): ...
+
+
+@dataclass
+class PageOfDatasetItems(PageOfItems[dict[str, Any]]): ...
+
+
+@dataclass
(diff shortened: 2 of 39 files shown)


src/apify_client/_resource_clients/actor_collection.py
```diff
@@ -55,9 +65,12 @@ def list(
         desc: bool | None = None,
         sort_by: Literal['created_at', 'last_run_started_at'] | None = 'created_at',
         timeout: Timeout = 'medium',
-    ) -> ListOfActors:
+    ) -> IterablePageOfActors:
```
Pijukatel, line 68: Naming parity with JS is not 100% as there it was possible to use generics, but the typing there allows more flexibility.

JS name:
`PaginatedIterator<ActorCollectionListItem>`

src/apify_client/_pagination_classes.py
```diff
@@ -0,0 +1,267 @@
...
+    def __await__(self) -> Generator[Any, Any, PageOfKeys]:
+        return self._awaitable_first_page.__await__()
+
+
+@dataclass
+class IterablePageOfActors(PageOfItems[ActorShort], IterableOf[ActorShort]): ...
+
+
+@dataclass
+class IterablePageOfActorsAsync(AwaitablePage[ActorShort], AsyncIterableOf[ActorShort]): ...
```
Pijukatel, line 147: I tried different approaches with generics, but this turned out to be the best approach:
- It truly describes all the attributes
- It has both static safety, and it benefits from run-time validation when these dataclasses are created from Pydantic models
- It has as little code duplication as possible due to the autogenerated composable init of `dataclass`
- It gives a clear overview of which endpoints are "weird" - not having the normal pagination metadata

tests/unit/test_client_pagination.py
```diff
@@ -0,0 +1,660 @@
...
+    'ActorVersionCollectionClient',
+}
+
+DATASET_CLIENTS = {'DatasetClient'}
+RQ_CLIENTS = {'RequestQueueClient'}
+KVS_CLIENTS = {'KeyValueStoreClient'}
+STORAGE_CLIENTS = DATASET_CLIENTS | RQ_CLIENTS | KVS_CLIENTS
+ALL_CLIENTS = COLLECTION_CLIENTS | NO_OPTIONS_CLIENTS | STORAGE_CLIENTS
+
+TEST_CASES = (
```
Pijukatel, line 391: Test setup is somewhat complex and this is due to test code being nearly identical for each endpoint, but different endpoints not being consistent in which pagination it supports (different params and different pagination metadata).

Through this parameter definition, over 200 test cases is generated based on just a few actual tests

tests/unit/test_client_pagination.py
```diff
@@ -0,0 +1,660 @@
...
+    'ListOfSchedulesResponse',
+    'ListOfTasksResponse',
+    'ListOfVersionsResponse',
+    'ListOfWebhooksResponse',
+)
+
+
+@pytest.fixture(autouse=True)
+def _relax_item_validation() -> Any:
+    """Relax only the element type of `items` on paginated list models for the test run.
```
Pijukatel, line 136: These tests are only about pagination logic and constructing each item correctly is totally out of scope, so in all Pydantic models the `items: list[whatever]` validation is replaced by `items: list[dict]`

docs/02_concepts/08_pagination.mdx
```diff
@@ -12,10 +12,10 @@ import ApiLink from '@site/src/components/ApiLink';
 
 import PaginationAsyncExample from '!!raw-loader!./code/08_pagination_async.py';
 import PaginationSyncExample from '!!raw-loader!./code/08_pagination_sync.py';
-
 import IterateItemsAsyncExample from '!!raw-loader!./code/08_iterate_items_async.py';
 import IterateItemsSyncExample from '!!raw-loader!./code/08_iterate_items_sync.py';
 
+
 Most methods named `list` or `list_something` in the Apify client return a <ApiLink to="class/ListPage">`ListPage`</ApiLink>  object. This object provides a consistent interface for working with paginated data and includes the following properties:
```
vdusek, line 19: Do we still have `ListPage`? Could you ask Claude to also review and update the docs?

vdusek (commented):
Sorry, but I'm really not sure about merging pagination and iteration into a single `list` method. These are two distinct use cases - you either want a page (with metadata), or you want to stream items - and choosing the right method is a way to express that intent.

I would say this also goes against the single responsibility principle - the method now serves two different concerns, which makes the end user interface worse in terms of DX.

If we look beyond the interface and go to the implementation, it becomes even worse. The current approach introduces a 14 `IterablePageOf...` classes plus mixins just to make the typing work. The return type carries both responsibilities;  the async variant is no longer a coroutine but a lazy awaitable-iterable wrapper, and each resource client ends up using a closure-over-callback pattern (where there used to be two simple lines before).

Could we keep `list()` for pagination and `iterate()` for iteration? It's a more intuitive interface; callers don't have to learn the quirks of a hybrid object, it will be more understandable for type checkers, and the implementation would be significantly simpler without this mess.

Pijukatel:
> Could we keep `list()` for pagination and `iterate()` for iteration? It's a more intuitive interface; callers don't have to learn the quirks of a hybrid object, it will be more understandable for type checkers, and the implementation would be significantly simpler without this mess.

If we agree on having it split into two methods in Python client, then it will be way simpler and cleaner internally.
@janbuchar , what do you think?

janbuchar:
> > Could we keep `list()` for pagination and `iterate()` for iteration? It's a more intuitive interface; callers don't have to learn the quirks of a hybrid object, it will be more understandable for type checkers, and the implementation would be significantly simpler without this mess.
> 
> If we agree on having it split into two methods in Python client, then it will be way simpler and cleaner internally. @janbuchar , what do you think?

Well, in https://github.com/apify/apify-client-js, we have a `list` (or `listItems` for dataset items) method that returns a `PaginatedIterator`. This object can either be `await`-ed to receive a single page of the results or `async`-iterated to process all items in all pages.

Our current doctrine is pretty much "1:1 parity unless we have a clearly superior approach that we want to eventually use in the other flavor as well". Also, we mostly tend to sacrifice maintainability if it means that the API will become easier to use.

Thus the one thing to consider from @vdusek response is if returning a "polymorphic" object that is both an async iterable and a simple promise is strictly worse DX than having two distinct methods for each list-shaped endpoint. If it is, then we should also decide if this is something that we'd like to adjust in the JS version or if there is a compelling argument for letting the JS version and the Python version diverge.

Pijukatel:
Here is the variant with explicit `iterate` methods 
https://github.com/apify/apify-client-python/pull/771

Pijukatel:
> Here is the variant with explicit `iterate` methods #771

In my view, (repeating some points already mentioned)
**1) `list` only solution**

**Positives:**
- Parity of clients
- We do not have to create an additional function with a nearly identical signature in each resource client

**Negatives:**
- Inconsistency in return values with other clients' methods. Other methods are returning Pydantic models, but combined `list/iter` can't return Pydantic model as the model already defines its own `__iter__`. So the combined `list/iter` returns some custom dataclass, which is kind of a clone of the autogenerated model.
- A bunch of handwritten classes that are duplicating code-generated Pydantic models  
- Internally ugly and complicated to be compliant with static type checkers
- Exotic interface - combining both list and iteration into one function might be confusing

****2) `list` and `iterate`** 
Positives:**
- Consistency with previous Python client version (it already had `iterate` methods for datasets items for example), so Python users will be familiar with this
- Fully reuses autogenerated models, no need to duplicate
- Consistent return value for different client methods - Pydantic models
- Single responsibility principle
- Simple interface 

**Negatives:**
- Inconsistency with JS client
- `iter` and `list` have nearly identical signatures - a lot of duplication


While 1:1 parity is important, I would say that following language specifics might be even more important. Python client users are mainly Python users. How much do they care about JS client interface parity? (The clients will offer the same functionality, just in a slightly different way. So I would lean towards the **2**)

janbuchar:
Thanks for the writeup @Pijukatel. I agree that this is not that big of a deal, so if this makes the autogeneration process simpler, it's probably worth it.

vdusek:
Perfect, it seems we have a consensus, so I'll close this one in behalf of https://github.com/apify/apify-client-python/pull/771.

PR #786, opened 2026-05-11
test: fix flaky `test_request_queue_unlock_requests`

vdusek:
## Summary

Fix [this flaky CI run](https://github.com/apify/apify-client-python/actions/runs/25495772865/job/74814967391?pr=742) of `test_request_queue_unlock_requests[sync]`, which failed with `assert 2 == 3` on `unlock_response.unlocked_count`.

**Root cause:** the polling loop called `list_and_lock_head` repeatedly, which has the side effect of locking items. Across multiple iterations (due to eventual consistency on `add_request`) the total set of actually-locked requests no longer matched the last response's `items` count, so `unlock_requests` could report a different count than the test expected.

**Fix:** separate the two concerns — poll the read-only `list_head` until all 5 added requests are visible, then call `list_and_lock_head(limit=3)` exactly once. This removes the cross-iteration locking ambiguity.

Applied the same treatment to `test_request_queue_list_and_lock_head`, which shared the same racy polling pattern.

### DIFF
--- a/tests/integration/test_request_queue.py
+++ b/tests/integration/test_request_queue.py
@@ -22,7 +22,32 @@ from apify_client._models import (
 
 if TYPE_CHECKING:
     from apify_client import ApifyClient, ApifyClientAsync
+    from apify_client._resource_clients.request_queue import RequestQueueClient, RequestQueueClientAsync
     from apify_client._typeddicts import RequestDict, RequestDraftDeleteDict, RequestDraftDict
+
+
+async def ensure_queue_is_populated(
+    rq_client: RequestQueueClient | RequestQueueClientAsync,
+    *,
+    expected_count: int,
+    is_async: bool,
+) -> None:
+    """Poll the queue until `expected_count` requests are visible.
+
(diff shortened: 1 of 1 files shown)


tests/integration/test_request_queue.py
```diff
@@ -411,17 +411,23 @@ async def test_request_queue_list_and_lock_head(client: ApifyClient | ApifyClien
...
-            if len(lock_response.items) == 3:
+            result = await maybe_await(rq_client.list_head(limit=5))
+            assert isinstance(result, RequestQueueHead)
+            head_response = result
+            if len(head_response.items) == 5:
                 break
 
-        assert lock_response is not None
+        assert head_response is not None
+        assert len(head_response.items) == 5
```
janbuchar: This could be a helper function instead (`ensure_queue_is_populated`?), calling `list_head` in a test focused on `list_and_lock_head` gives off bad vibes.

janbuchar (approved):

Issue #780, opened 2026-05-05
docs: `ApiLink` generates incorrect URL for versioned docs


barjin:
`ApiLink` always resolves to `/api/class/<ClassName>` regardless of the active Docusaurus version. For the next version, it should resolve to `/api/next/class/<ClassName>`.

This causes the docs CI check to fail when a PR introduces new public classes, since they don't exist in the stable API docs yet.

Parity tracking for https://github.com/apify/crawlee-python/issues/1867

---

Note - the right way to go about this is likely moving `ApiLink` to a shared theme. Currently, there are small differences between the `ApiLink` implementations in the projects, though, so it's not straightforward.

PR #790, opened 2026-05-13
docs: bump `docusaurus-plugin-typedoc-api` version

barjin:
Bumps the API reference rendering plugin version. Related to #780


vdusek (approved):
thanks, just let's use "docs" so that it gets updated 🙂

PR #793, opened 2026-05-14
feat: accept camelCase keys in input TypedDicts

### LINKED ISSUE #756: Consider accepting both snake_case and camelCase values for dicts and Literals
## Context

Spun out of a review discussion on #738: https://github.com/apify/apify-client-python/pull/738#discussion_r3111857207

> I think there is a case for accepting camelCase here instead - it is exactly what the API accepts. On the other hand, snake_case fits in better with the rest of the Python code and works better for the "I just don't want to look up what model I need to import" use case for TypedDicts.

The same tension already exists today for `Literal` values used in resource-client methods, e.g.:

```python
sort_by: Literal['created_at', 'last_run_started_at'] | None = 'created_at'
```

The Apify API expects these as camelCase (`createdAt`, `lastRunStartedAt`), while the Python client exposes them in snake_case to match Python conventions. The same question now arises for the newly generated input-side `TypedDict`s from #738 (field names in dicts passed to client methods).

## Proposal

Consider — and, if agreed, implement — accepting **both** `snake_case` and `camelCase` values in:

1. **`TypedDict` inputs** introduced in #738 (`_typeddicts_generated.py`, `_typeddicts.py`) — allow users to pass either naming style as keys.
2. **`Literal` arguments** already present across resource clients (e.g. `sort_by`, pricing-model literals, webhook event-type literals, etc.) — accept both forms.

### Trade-offs to weigh

- **Pro snake_case only:** consistent with Pydantic models and the rest of the Python code; idiomatic for Python users.
- **Pro camelCase only:** matches the API 1:1 — users who already know the API don't need a translation table, which is especially valuable for the "I don't want to import the Pydantic model, just pass a dict" use case.
- **Pro accepting both:** maximum ergonomics, no wrong answer for the user; cost is implementation complexity (normalization layer) and documentation/typing clarity (both forms must be reflected in the type hints so the type checker accepts them).

## Scope

- Decide on the policy (one style only vs. both accepted).
- If accepting both: define normalization boundary (input translation to API payload), update generation scripts (`scripts/postprocess_generated_models.py`) and the hand-written literals in resource clients, and ensure type checkers (`ty`) see both forms.
- Update docs/examples accordingly.

cc @janbuchar

vdusek:
## Summary

Synthesize a `<Name>CamelDict` sibling for every input TypedDict so users can pass API-shaped dicts (e.g. `{'uniqueKey': ..., 'retryCount': ...}`) and still satisfy the type checker. The Pydantic alias map extracted from `_models.py` is the source of truth for the camelCase names; nested TypedDict references in cloned annotations are rewired to their camel variants.

Resource-client signatures in `_resource_clients/request_queue.py` and the `WebhooksList` union in `_types.py` are widened to accept either casing alongside the Pydantic model.

Runtime behavior is unchanged — Pydantic's `populate_by_name=True` already validates both casings.

Closes #756.


janbuchar:
Well, the code looks pretty much the way I'd expect it to look - not great, not terrible :slightly_smiling_face: Did you test what the DX feels like? I.e., type checking, completions, ...?

vdusek:
@janbuchar The type checking works perfectly (missing required field, mixing snake+camel, invalid literal value, unknown extra keys, ...). Autocomplete inside inline typeddicts literals (if you mean this) doesn't work, at least for me in vscode. AFAIK, it has never worked for me there (vscode + typeddicts), even with Pylance LSP. Maybe current limitation of the dev tooling, or I have something wrong in the settings :slightly_smiling_face:.

janbuchar (approved):
Nice, LGTM

PR #794, opened 2026-05-15
refactor: Consolidate public interface for v3

### LINKED ISSUE #744: Revise public interface of types before v3 release
Follow-up from discussion: https://github.com/apify/apify-client-python/pull/738#discussion_r3111526151

Before the v3 release, do a final revision of the public interface related to generated types — models, `TypedDict`s, enums, `Literal`s, etc. In particular, the `TypedDict`s should be exposed as public.

vdusek:
### Description

Consolidates the public interface for the v3 release:
  - Root `apify_client` namespace now only exposes `ApifyClient`, `ApifyClientAsync`, and `__version__`. This is a go-to for everyone.
  - HTTP client abstractions moved to `http_clients` submodule (`HttpClient`, `HttpClientAsync`, `HttpResponse`, `ImpitHttpClient`, `ImpitHttpClientAsync`). Only for people who want to replace the default HTTP client.
- New public `types.py` module with `Timeout`, `WebhooksList`, and `JsonSerializable` type aliases (can be used, e.g., for a custom HTTP client).
- Add missing `__all__` to `errors.py`.
- Update v3 upgrade guide and doc snippets to reflect new import paths.
- Not breaking, because HTTP clients were not part of v2.5.

I would prefer to keep `_models` and `_typeddicts` private. They should not be considered safe to import, since exposing them publicly would potentially restrict future API changes. For typed dicts, users do not need to import them to pass inputs to resource client methods. Resource clients, which are also private and users should be able to get to them only via `ApifyClient` (and its async version), may return Pydantic models, but those models should not be constructed directly in user code. There may be niche cases where users want to import models or typed dicts for explicit typing or casts, but I would rather keep those available only on the user's own responsibility - we cannot guarantee that the types will not change. This is also a status quo in the current v2.5.

This is my stand, let me know yours @janbuchar.

### Issue

Closes: #744


janbuchar:
I'm not sure if `apify_client.types` is actually necessary to develop a new Client, but otherwise I completely agree with the changes, thanks, well done!

janbuchar (approved):

vdusek:
> I'm not sure if apify_client.types is actually necessary to develop a new Client, but otherwise I completely agree with the changes, thanks, well done!

Well, probably only `Timeout` is needed, so wondering... Maybe we shall keep `_types.py` private and add a new module `types.py` exporting public "types"?

janbuchar:
> > I'm not sure if apify_client.types is actually necessary to develop a new Client, but otherwise I completely agree with the changes, thanks, well done!
> 
> Well, probably only `Timeout` is needed, so wondering... Maybe we shall keep `_types.py` private and add a new module `types.py` exporting public "types"?

Yeah, that sounds good to me. Maybe we should think about the naming some more though

PR #828, opened 2026-05-26
ci: Allow manually dispatching the master Checks workflow

vdusek:
## Summary

Adds `workflow_dispatch` to `CI (master)` so the `Checks` suite can be re-run manually on the same commit when the push-triggered run is lost — e.g. during a GitHub outage. The `doc_release` / `beta_release` jobs are gated on `github.event_name == 'push'` so manual dispatches only re-run the checks and never re-publish.

The redundant `workflow_dispatch` trigger on `_checks.yaml` is removed: dispatching it directly produced check-run names without the `Checks /` prefix, so the `wait-for-checks` gate in the manual release workflows wouldn't recognize them anyway.

Mirrors apify/crawlee-python#1917.

## How to use

When a push-triggered Checks run on `master` is missing, go to **Actions → CI (master) → Run workflow** on `master`. Once it succeeds, the manual release workflows' `wait-for-checks` gate will be satisfied.

### DIFF
--- a/.github/workflows/on_master.yaml
+++ b/.github/workflows/on_master.yaml
@@ -7,6 +7,16 @@ on:
     tags-ignore:
       - "**" # Ignore all tags to avoid duplicate executions triggered by tag pushes.
 
+  # Allow re-running the Checks manually. Release jobs below are gated on `github.event_name == 'push'`
+  # so they never fire on manual dispatch.
+  workflow_dispatch:
+    inputs:
+      run_tests:
+        description: Whether to run the test suites (unit, integration).
+        required: false
+        type: boolean
+        default: true
+
 permissions:
   contents: read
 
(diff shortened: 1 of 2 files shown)


.github/workflows/on_master.yaml
```diff
@@ -15,13 +25,14 @@ jobs:
...
+      # On push: skip the test suites for docs-only commits — they don't change runtime behavior.
+      # On manual dispatch: honor the input.
+      run_tests: ${{ github.event_name == 'push' && !startsWith(github.event.head_commit.message, 'docs') || github.event_name == 'workflow_dispatch' && inputs.run_tests }}
     secrets: inherit
 
   doc_release:
-    # Skip this for non-"docs" commits.
-    if: startsWith(github.event.head_commit.message, 'docs')
+    # Only on push, and only for "docs" commits.
+    if: github.event_name == 'push' && startsWith(github.event.head_commit.message, 'docs')
```
janbuchar, line 35: If someone wants to release the docs or a beta manually, I suppose they should just run that workflow manually instead of the whole on_master pipeline?

janbuchar (approved):

.github/workflows/on_master.yaml
vdusek: Exactly.

Manual execution of `on_master` should there be only for the checks.

PR #836, opened 2026-05-29
test: Fix flaky webhook dispatch pagination test

vdusek:
## Summary

`test_webhook_dispatch_list_pagination` was flaky: with `desc=True` (newest first), new dispatches fired by parallel xdist workers landed between the `offset=0` and `offset=5` `list()` calls, shifting `page.items[0]` so the inequality assertion failed.

Switched to `desc=False` (oldest first) so new dispatches land at the tail and offsets 0–9 stay put, and updated the sort-order assertion to ascending.

## Failing CI run

https://github.com/apify/apify-client-python/actions/runs/26616401723/job/78432968995


vdusek:
closing in favor of https://github.com/apify/apify-client-python/pull/843

PR #837, opened 2026-05-29
test: unify polling helpers with apify-sdk-python

vdusek:
Adds generic `call_with_exp_backoff` and `poll_until_condition` test helpers with signatures shared across `apify-sdk-python`, `crawlee-python`, and `apify-client-python`. Both accept sync or async callables (via a small `_maybe_await` helper, typed with overloads), which fits this repo's sync/async client duality.

`collect_iterate_until_present` is rebuilt on top of `poll_until_condition`: the iterate/drain logic stays, but the polling loop is delegated to the shared helper. The contract is unchanged — the last collected list is returned so callers keep asserting with their own failure messages.

Lint, type-check, and test collection pass; integration tests require a platform token and were not run locally.

### DIFF
--- a/tests/integration/_utils.py
+++ b/tests/integration/_utils.py
@@ -1,17 +1,21 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
+import logging
 import secrets
 import string
 import time
 from collections.abc import AsyncIterator, Iterator
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


vdusek:
closing in favor of https://github.com/apify/apify-client-python/pull/844

PR #842, opened 2026-06-02
ci: Approve pnpm build scripts for docs build

vdusek:
The recent bump to pnpm v11 (#841) turned ignored package build scripts into a hard error, breaking the docs check:

```
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: @apify/ui-icons, @swc/core, core-js, unrs-resolver
```

This explicitly approves those packages' build scripts via an `allowBuilds` section in `website/pnpm-workspace.yaml` (pnpm 11's mechanism for opting dependencies into running install scripts).

Verified locally that `pnpm install` and `uv run poe build-docs` both succeed.

### DIFF
--- a/website/pnpm-workspace.yaml
+++ b/website/pnpm-workspace.yaml
@@ -11,4 +11,5 @@ publicHoistPattern:
 allowBuilds:
   '@apify/ui-icons': true
   '@swc/core': true
+  core-js: false
   unrs-resolver: true
--- a/website/pnpm-workspace.yaml
+++ b/website/pnpm-workspace.yaml
@@ -11,5 +11,4 @@ publicHoistPattern:
 allowBuilds:
   '@apify/ui-icons': true
   '@swc/core': true
-  core-js: true
   unrs-resolver: true
--- a/website/pnpm-workspace.yaml
+++ b/website/pnpm-workspace.yaml
@@ -8,3 +8,8 @@ linkWorkspacePackages: true
 preferWorkspacePackages: true
 publicHoistPattern:
     - "*"
+allowBuilds:
+  '@apify/ui-icons': true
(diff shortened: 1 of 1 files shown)


vdusek:
@B4nan, could you please check it? If this is the right way, I'll do it in the SDK & Crawlee as well.

website/pnpm-workspace.yaml
```diff
@@ -8,3 +8,8 @@ linkWorkspacePackages: true
 preferWorkspacePackages: true
 publicHoistPattern:
     - "*"
+allowBuilds:
+  '@apify/ui-icons': true
+  '@swc/core': true
+  core-js: true
```
B4nan: you surely dont need this one, he only uses the script to spam with some sponsoring message

B4nan (approved):
lets drop core-js, otherwise it looks good.

website/pnpm-workspace.yaml
vdusek: Well, it seems it is needed, see https://github.com/apify/apify-client-python/actions/runs/26870774068/job/79245262125

B4nan: This must be coming from the way you run it via `uv` or something similar. The whole point of pnpm is to block unwanted scripts, it shouldn't fail just because you don't allow something. And the one from core-js is an unwanted script.

https://github.com/zloirock/core-js/blob/master/packages/core-js/postinstall.js

vdusek: Okay, so we can explicitly disallow it, and it looks like everything is good now.

B4nan: Right, I checked the other repos and this is what we do there too.

PR #844, opened 2026-06-02
test: deflake integration tests by polling instead of fixed sleeps

vdusek:
## Summary

Started as a fix for [this flaky CI run](https://github.com/apify/apify-client-python/actions/runs/26818252886/job/79065522664) of `test_request_queue_unlock_requests[sync]` (`assert 2 == 3` on `unlocked_count`, caused by replication lag between `list_and_lock_head` and `unlock_requests`), and grew into deflaking the integration test suite as a whole.

**Changes:**

- Add a `poll_until_condition` helper to `tests/integration/_utils.py`: polls a sync-or-async callable at a constant interval until a condition holds or a wall-clock timeout expires. An optional `backoff_factor` multiplies the interval after each poll, for highly variable waits (e.g. Actor run container startup) where a growing delay covers a long timeout with few calls.
- Fix the flaky unlock test by polling `list_head` until the locked IDs disappear from the queue head before unlocking.
- Replace all hand-rolled `for _ in range(5): sleep(1); read; break` polling loops (10×) and single fixed `sleep(1)` waits (26×) across the request queue, dataset, key-value store, and run integration tests with `poll_until_condition`. This makes the tests both faster on the happy path (no unconditional sleep) and more robust under load (polls until the timeout instead of hoping 1 s is enough).
- Generalize `maybe_await` to accept any awaitable.

The three `iterate_keys` sleeps in the KVS tests are intentionally left as-is: draining an iterator per attempt wants attempt-count semantics (like `collect_iterate_until_present`), not a wall-clock deadline.

Follow-up to #786.


tests/integration/test_run.py
```diff
@@ -329,9 +334,13 @@ async def test_run_reboot(client: ApifyClient | ApifyClientAsync, *, is_async: b
...
     try:
-        # Wait a bit and check if the run is still running
-        await maybe_sleep(1, is_async=is_async)
-        current_run = await maybe_await(run_client.get())
+        # Wait until the run starts (leaves READY), with exponential backoff: container startup time varies widely
+        async def get_run() -> Run | None:
+            return await maybe_await(run_client.get())
+
+        current_run = await poll_until_condition(
+            get_run, lambda run: isinstance(run, Run) and run.status != 'READY', timeout=30, backoff_factor=2
```
janbuchar, line 342: Is it necessary to accept _two_ callbacks in `poll_until_condition`? Can't it just require a function that returns `bool`? 

Also I keep thinking that this whole thing could be made a bit more readable with decorators, but I haven't actually come up with anything usable 😁

janbuchar (approved):

tests/integration/test_run.py
vdusek: > Is it necessary to accept two callbacks in poll_until_condition? Can't it just require a function that returns bool?

Well, I would say it is, because some callers need the value that was polled, not just a yes/no. Basically, `poll_until_condition` returns the last result of `fn`, and the condition only decides when to stop polling.

vdusek: So this is the current state:

```python
# Poll until all items are visible (eventual consistency)
async def get_items() -> DatasetItemsPage:
    page = await maybe_await(dataset_client.list_items())
    assert isinstance(page, DatasetItemsPage)
    return page

await poll_until_condition(get_items, lambda page: len(page.items) == 10)
```

And this would be your suggested approach:

```python
# Poll until all items are visible (eventual consistency)
async def get_items() -> DatasetItemsPage | None:
    page = await maybe_await(dataset_client.list_items())
    assert isinstance(page, DatasetItemsPage)
    return page if len(page.items) == 10 else None

await poll_until(get_items)
```

If I am not mistaken.

So WDYT? I'm not 100% sure, but the current state seems to me a bit intuitive from the caller's perspective. Call this until this happens, then return me the result.

janbuchar: Yeah, I think you got it right. I like that the second one accepts just one function, but it's just a test helper, so feel free to keep it this way.

PR #858, opened 2026-06-16
chore: use alias_generator on generated models instead of per-field aliases

### LINKED ISSUE #852: Use `alias_generator=to_camel` on generated Pydantic models instead of per-field aliases
### Background

We're on Pydantic ≥2.11. `src/apify_client/_models.py` is generated by datamodel-codegen: ~190 models with ~606 explicit `Field(alias='camelCase')` entries, e.g.:

```python
source_type: Annotated[VersionSourceType | None, Field(alias='sourceType')] = None
is_public: Annotated[bool, Field(alias='isPublic', examples=[False])] = None
```

All models already share `ConfigDict(extra='allow', populate_by_name=True)`.

### Proposal

Drive the codegen so each model's config carries `alias_generator=to_camel`, and emit an explicit alias only where the conversion is irregular:

```python
from pydantic.alias_generators import to_camel

model_config = ConfigDict(extra='allow', populate_by_name=True, alias_generator=to_camel)
```

### Where it makes sense

Because this file is generated, the work lives in the generation + postprocess pipeline rather than in hand edits:

- Irregular conversions still need an explicit alias. The existing override in `datamodel_codegen_aliases.json` (`gitHubGistUrl` → `github_gist_url`) is a good example: `to_camel('github_gist_url')` yields `githubGistUrl`, not `gitHubGistUrl`, so such overrides must win over the generator.
- `scripts/postprocess_generated_models.py` builds its alias map by AST-walking `Field(alias=...)` to synthesize the camelCase `*Dict` TypedDict variants. If aliases stop being emitted per field, that step must instead derive them via `to_camel` plus the override map. This is the main thing to verify.

### Scope

Refactor only, smaller generated output. The wire format must stay identical, guarded by tests.

vdusek:
The codegen postprocess now adds `alias_generator=to_camel` to every model's `ConfigDict` and emits an explicit `Field(alias=...)` only where the camelCase conversion is irregular (all-caps usage keys, `gitHubGistUrl`, `schema`, ...). This drops the explicit per-field aliases in `_models.py` from 623 to 45 and shrinks the file from 3852 to 3077 lines.

The wire format is unchanged, guarded by tests: Pydantic lets an explicit alias override the generator, and `populate_by_name=True` keeps snake_case input working. The before/after runtime alias contract is byte-identical across all 212 models / 1060 fields, and `_typeddicts.py` / `_literals.py` regenerate unchanged (the camel `*Dict` synthesis now derives names via `to_camel` plus the irregular-alias overrides).

datamodel-codegen has no native option for this (its `--no-alias` would drop the irregular aliases too), so the work lives in `scripts/postprocess_generated_models.py`.

Closes #852


vdusek:
I opened a feature request to https://github.com/koxudaxi/datamodel-code-generator/issues/3419 , so let's see.

Pijukatel (approved):

PR #866, opened 2026-06-18
fix: Only log any error that happens during log redirection

### LINKED ISSUE #864: Update client to handle new errors from impit
Impit might start throwing new errors that were previously silently ignored. Update the client so that especially log redirection never throws an exception, only warnings.

For context see:
https://github.com/apify/impit/pull/482#issuecomment-4718449924

Pijukatel:
# Description
- Never raise an error on a log redirection problem. 
(This was hidden by [a bug in impit, which is going to be fixed soon](https://github.com/apify/impit/pull/482) and it would lead to similar issue observed in JS client https://github.com/apify/apify-client-js/issues/873)
- Use `~` on `impit` and pin to the latest version `0.12.0` . Zero versioned releases can have breaking changes even in minor versions. Bumping impit version to new minor version should be an intentional action. (Expected release of impit `0.13.0` would break client without this change)

# Issues
Closes: https://github.com/apify/apify-client-python/issues/864


barjin (approved):
lgtm, tested as part of triaging https://github.com/apify/impit/pull/482 👍 

Thank you, @Pijukatel !

PR #887, opened 2026-06-24
chore(ci): Update renovate schedules

Pijukatel:
- Replace deprecated `breejslater` syntax with `cron` syntax
https://docs.renovatebot.com/key-concepts/scheduling/#deprecated-breejslater-syntax
- Use these triggers, limit the window to avoid spam:
  - Every **Tuesday - Friday** between **1:00 and 2:00** to update **dependencies**
  - Every **Monday** between **1:00 and 2:00** to update **lockfiles** 
- Do not run both jobs on the same day.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,7 +9,7 @@
         "enabled": true,
         "automerge": true,
         "automergeType": "pr",
-        "schedule": ["* 1-3 * * 1"]
+        "schedule": ["* 1-2 * * 1"]
     },
     "packageRules": [
         {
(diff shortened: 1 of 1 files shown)


vdusek (commented):
Could we use between 1 AM and 3 AM for both cases?

Pijukatel:
> Could we use between 1 AM and 3 AM for both cases?

I was just wondering how those two jobs will interact with each other. Just to reduce the risk of conflicting PRs or bumping the lock file twice a day and creating to much spam for us, maybe they could both be from 1 AM - 3 AM, but on Monday, it would be just lock file maintenance? And dependency bump Tuesday-Friday?

vdusek (approved):

PR #927, opened 2026-07-08
feat: Add request body compression with optional brotli

### LINKED ISSUE #942: Make brotli compression quality configurable
PR #927 added optional brotli request-body compression. The brotli quality is currently **hardcoded to 6** in [`src/apify_client/http_clients/_base.py`](https://github.com/apify/apify-client-python/blob/master/src/apify_client/http_clients/_base.py):

```python
_brotli_compress = functools.partial(brotli.compress, quality=6)
```

Quality 6 is a reasonable default (better ratio than gzip, much lower CPU than brotli's max quality 11), and it's documented as such in `docs/02_concepts/13_compression.mdx`. But it's not tunable: users who want maximum compression (quality 11) for bandwidth-bound workloads, or a lower quality to save CPU, currently have no way to change it.

## Proposal

Expose the compression quality as a client-level configuration option instead of relying only on the hardcoded value. Points to decide during design:

- A client parameter (e.g. on `ApifyClient` / the HTTP client) to override the brotli quality, defaulting to the current `6`.
- Whether to also make the gzip compression level configurable for symmetry.
- Whether to let users force a specific algorithm rather than relying purely on import-based auto-detection (`br` when `brotli` is importable, `gzip` otherwise) — the current behavior flips to brotli whenever the `brotli` package is present in the environment, even if it was pulled in transitively.

## Context

- PR that introduced compression: #927
- Discussion on the PR: https://github.com/apify/apify-client-python/pull/927
- Related issue (min-size threshold, separate concern): #934

mixalturek:
- https://github.com/apify/apify-core/pull/28971 added support for brotli compression to BE.
- Pros: higher compression, cons: more CPU intensive.
- The brotli dependencies are defined as optional, one has to explicitly enable it and choose one. If none is available, the code falls back to gzip.
- JS client doesn't compress requests that are too small, this Python client compresses just everything. No change done, I only noticed and stating it.

Closes: #942

### Related PRs
- https://github.com/apify/apify-core/pull/28971
- https://github.com/apify/apify-docs/pull/2750
- https://github.com/apify/apify-client-js/pull/962
- https://github.com/apify/apify-client-python/pull/927
- https://github.com/apify/apify-sdk-python/pull/1031


src/apify_client/http_clients/_base.py
```diff
@@ -203,22 +218,30 @@ def _prepare_request_call(
...
             data = jsonlib.dumps(json, ensure_ascii=False, allow_nan=False, default=str).encode('utf-8')
             headers['Content-Type'] = 'application/json'
 
         if isinstance(data, (str, bytes, bytearray)):
             if isinstance(data, str):
                 data = data.encode('utf-8')
-            data = gzip.compress(data)
-            headers['Content-Encoding'] = 'gzip'
+            if _brotli_compress is not None:
+                data = _brotli_compress(data)
```
vdusek: From Claude:

> Important: a `bytearray` body reaches `_brotli_compress` unconverted (only `str` is encoded above), but `brotlicffi.compress` — the variant the `brotli` extra installs on non-CPython — rejects `bytearray` with `TypeError`. The old `gzip.compress` accepted it, so this is a regression, reachable via `data=bytearray(...)` or `key_value_store.set_record(k, bytearray(...))`. `bytes(x)` is a no-op for `bytes` and cheap for `bytearray`:

src/apify_client/http_clients/_base.py
```diff
@@ -23,10 +23,25 @@
...
 
     from apify_client.types import JsonSerializable, Timeout
 
+_brotli_compress: Callable[[bytes | bytearray], bytes] | None = None
+
+if not TYPE_CHECKING:
+    try:
+        import brotli
+
+        _brotli_compress = brotli.compress
```
vdusek: From Claude, we should consider this:

> Note: `brotli.compress` / `brotlicffi.compress` default to `quality=11`, the maximum and slowest level, applied to every request body on the hot path (e.g. large `push_items` batches). For a marginal ratio gain over quality 4-6 this can be several times slower than gzip. Consider binding a lower quality at both sites (this line and line 41), e.g. `functools.partial(brotli.compress, quality=5)`.

tests/unit/test_http_clients.py
```diff
@@ -452,3 +465,41 @@ def test_build_url_with_params_mixed() -> None:
     assert 'tags=a' in url
     assert 'tags=b' in url
     assert 'name=test' in url
+
+
+def test_prepare_request_call_brotli_compression(monkeypatch: pytest.MonkeyPatch) -> None:
+    """When a brotli compressor is available, request body uses brotli (Content-Encoding: br)."""
+    brotlicffi = pytest.importorskip('brotlicffi')
```
vdusek: Thanks to this, the tests are being skipped in the CI now. Is it intentional?

tests/unit/test_run_charge.py
```diff
@@ -4,6 +4,25 @@
 import json
 from typing import TYPE_CHECKING
 
+try:
+    import brotlicffi as _brotli_mod
```
vdusek: This brotli-detection ladder is now the third near-identical copy (see `_base.py:33` and `test_http_clients.py:16`), and on top of that, its import order is inverted.

Would it be possible to come up with something better?

pyproject.toml
```diff
@@ -203,7 +209,7 @@ exclude = ["website/versioned_docs"]
 unused-ignore-comment = "error"
 
 [[tool.ty.overrides]]
-include = ["docs/**/*.py", "website/**/*.py"]
+include = ["docs/**/*.py", "website/**/*.py", "tests/unit/test_http_clients.py", "tests/unit/test_run_charge.py"]
```
vdusek: This is wrong. Add new packages to the dev dependency group.

README.md
```diff
@@ -48,6 +48,15 @@
 
     or any other Python package manager that consumes PyPI.
 
+    To enable brotli request-body compression (better than gzip, especially for large payloads), install the optional extra:
```
vdusek: I'm just wondering whether this should really be the only way to configure the compression method (only via installation of extra), or whether we should introduce a new flag on the `ApifyClient` base class instead.

@Pijukatel WDYT?

pyproject.toml
```diff
@@ -30,6 +30,12 @@ dependencies = [
     "pydantic[email]>=2.11.0",
 ]
 
+[project.optional-dependencies]
+brotli = [
+    "brotli>=1.0.9; platform_python_implementation == 'CPython'",
```
vdusek: Just curious, why exactly `>= 1.0.9`?

tests/unit/test_run_charge.py
```diff
@@ -4,6 +4,25 @@
...
+
+    def _decompress(data: bytes) -> bytes:
+        return _brotli_mod.decompress(data)
+
+except ImportError:
+    try:
+        import brotli as _brotli_mod  # type: ignore[no-redef]
```
vdusek: This is a `mypy` diagnostic code, we use `ty` in all our projects.

pyproject.toml
```diff
@@ -30,6 +30,12 @@ dependencies = [
...
+    "brotlicffi>=1.0.9; platform_python_implementation != 'CPython'",
```
vdusek: Do we now support other platforms than CPython?

vdusek (changes requested):

vdusek:
This has a significant impact. @Pijukatel, please check it as well.

pyproject.toml
mixalturek: Only CPython is supported; this should simplify the code significantly.

```
⏺ That's a definitive answer. Every single impit wheel is cp* (CPython only) — no pp* PyPy wheels, no py3-none-any pure-Python wheel. The source tarball exists but building it from source requires a Rust toolchain, and the resulting binary would still only target CPython unless impit's Cargo.toml is configured   
  for PyPy via PyO3's abi3 or PyPy support.                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                          
  Summary: impit is the hard blocker. It's a mandatory, non-optional dependency, and it only ships CPython wheels. On PyPy, installation would fail unless someone adds PyPy wheel support to impit upstream. Everything else in the repo is PyPy-compatible:                                                             
                                                                                                                                                                                                                                                                                                                          
  - pydantic-core (Rust) — does publish pp311-pypy311 wheels ✓                                                                                                                                                                                                                                                            
  - brotli/brotlicffi — the platform_python_implementation markers already handle PyPy correctly ✓                                                                                                                                                                                                                        
  - colorama, more-itertools — pure Python ✓                                                                                                                                                                                                                                                                              
  - The client code itself — no CPython-specific APIs ✓                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                          
  If PyPy support were a goal, it would require either getting impit to publish PyPy wheels, or adding a PyPy-compatible HTTP client backend (e.g., httpx) behind the with_custom_http_client() hook and making impit an optional dependency.
```

pyproject.toml
mixalturek: I asked Claude as well :-D

```
  1.0.9 was effectively the only version on PyPI for years.                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                            
  The brotli Python package (Google's C-extension wrapper) published 1.0.9 in August 2020 and then went dormant — a 1.2.0 existed on GitHub but was never published to PyPI until late 2025. So when the feature was implemented in this project, >=1.0.9 was essentially "use the only available version."                 
                                                                                                                                                                                                                                                                                                                            
  The project's lock file confirms this: it currently resolves to brotli 1.2.0 (released 2025-11-05), meaning the constraint >=1.0.9 still holds and new versions are picked up automatically.                                                                                                                              
                                                                                                                                                                                                                                                                                                                            
  Two secondary reasons 1.0.9 is a good floor:                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                            
  - It includes the fix for CVE-2020-8927 (a buffer overflow in the encoder for inputs >2 GiB, present in 1.0.8).                                                                                                                                                                                                           
  - brotlicffi (originally also in the optional deps, later removed in 851a2df) uses the same version numbering scheme — brotlicffi 1.0.9.x tracks the brotli C library 1.0.9. Keeping both aligned at >=1.0.9 made the constraint symmetric.
```

https://pypi.org/project/brotli/#history

src/apify_client/http_clients/_base.py
mixalturek: Solved by `brotlicffi` removal, `brotli` dependency handles it without any issue, a new test added.

tests/unit/test_run_charge.py
mixalturek: Fixed by adding it to dev dependencies and using it directly in tests.

mixalturek:
@vdusek Nearly all comments are fixed now, thanks again for your great review.

One more thing: I forgot to mention that the JS client skips compression if the payload size is below 1024 B. Do we want to apply it to the Python client as well?

pyproject.toml
vdusek: Makes sense, thanks :slightly_smiling_face:

pyproject.toml
vdusek: Thanks.

vdusek:
> One more thing: I forgot to mention that the JS client skips compression if the payload size is below 1024 B. Do we want to apply it to the Python client as well?

Thanks for noticing. I opened an issue for that: https://github.com/apify/apify-client-python/issues/934. Let's resolve it in a separate PR so it is documented in the changelog.

src/apify_client/http_clients/_base.py
Pijukatel: I would challenge this comment. Not sure whether there is any point in an optimization debate without numbers, though.

But what is the point of compression? Optimization of transfer speed, since it is usually many times slower than the CPU time we spend on better compression. 

So I would suggest starting with the default and changing it only if we have evidence that it is not good enough.

README.md
Pijukatel: Is there a use case where someone makes an effort to install brotli and wants to use gzip?

I would start without the option for now and add it only if it proves to be a valid and common use case. (Easier to extend than remove)

Pijukatel (approved):
Not sure about the non-default compression quality , but it is not blocking for me.

src/apify_client/http_clients/_base.py
vdusek: Here are some numbers:

image

And some additional reasoning:

> On "start with the default": q11 is not a neutral default for this use case. It's the extreme designed for static assets: compress once, serve many times, so the CPU cost is amortized. For per-request, on-the-fly compression, where each payload is compressed once and sent once, the ecosystem default is q4–6: nginx ngx_brotli uses 6, and Cloudflare dynamic compression uses 4–5. So q6 is “starting with a sensible default”; q11 would be the unusual choice that needs justification.

So my take is that for projects like a general API client, a high compression level would only be useful and pay off in cases with very poor bandwidth. But that is more of an argument for making this configurable, rather than using it as the default.

For the default, we should stay somewhere around what nginx and Cloudflare use.

vdusek (commented):
I removed `brotli` from the dev dependency group. Sorry, my previous comment wasn't entirely correct. We run `uv sync --all-extras` everywhere in CI, so there's no need to include it there. The initial mypy/ty override ignores confused me, but they were probably added based on your local setup, where you didn't have the `brotli` extra installed.

Other than that, one more thing: the documentation 🙂.

Please extend the installation section in the docs here: https://docs.apify.com/api/client/python/docs#installation. You can use a similar section to the one you added to the README. 

Also, please create a new concept page for compression. Something like: explaining HTTP compression, how to switch between compression methods, and the pros and cons of each approach.

src/apify_client/http_clients/_base.py
Pijukatel: Thanks for the data. Since the q6 settings is improvement over gzip both in terms of speed and compression, it is truly a good candidate.

src/apify_client/http_clients/_base.py
```diff
@@ -23,10 +24,20 @@
...
-    from collections.abc import AsyncIterator, Iterator, Mapping
+    from collections.abc import AsyncIterator, Callable, Iterator, Mapping
 
     from apify_client.types import JsonSerializable, Timeout
 
+_brotli_compress: Callable[[bytes | bytearray], bytes] | None = None
+
+if not TYPE_CHECKING:
+    try:
+        import brotli
```
vdusek: I'm still thinking about this, and I'm not sure I like the current approach. Brotli is enabled only based on whether `brotli` can be imported, rather than through a configurable option.

And this brings some issues, for example, if another dependency installs `brotli` transitively, the client will silently start using `br` even if the `[brotli]` extra was never selected. There is also no runtime option to switch back to gzip.

I'd prefer an explicit client-side option for the encoding type (probably on the facade `ApifyClient` and `ApifyClientAsync` classes). That would also make it easy to expose the brotli quality level configuration, with `6` as the default.

You can also take inspiration from how we handle the missing extras in the [crawlee-python](https://github.com/apify/crawlee-python).

tests/unit/test_http_clients.py
```diff
@@ -5,6 +5,7 @@
 from typing import Any
 from unittest.mock import Mock
 
+import brotli
```
vdusek, line 9: If I'm not mistaken, we've switched all tests from gzip to brotli, with gzip covered only in the final test to verify the fallback behavior.

I'd prefer to see these tests parametrized. Any test that involves encoding should run against both brotli and gzip so that both options are covered consistently.

docs/02_concepts/13_compression.mdx
```diff
@@ -0,0 +1,42 @@
+---
+id: compression
+title: Request body compression
+description: The client compresses every request body automatically, using brotli when available or gzip as a fallback.
+---
+
+The Apify client compresses every request body before sending it to the API. This reduces the amount of data transferred over the network, resulting in faster requests and lower bandwidth usage — especially for large payloads such as Actor inputs, dataset uploads, or key-value store records.
```
vdusek: Doc writing style (@szaganek will find more 😄):

> Per our docs style, avoid em-dashes gluing clauses (here and line 28) and prefer contractions in the tip on line 41 ("you're", "isn't"); splitting into shorter sentences reads better.

src/apify_client/http_clients/_base.py
mixalturek: I tried to generate code to see what it could look like; see the last commit.

Frankly, I am not a fan of this approach; it's quite a lot of code and documentation for too low value. I personally expect this feature and possibility will be used only very rarely, and everyone will stick with the defaults. Both options work comparably well, and people are lazy :-) I just feel this is an overengineering; let's keep it simple and wait for a feature request before "messing" the code and interfaces. In general, it's simple to add, but difficult to remove. 

I would either prefer to let the library be opinionated or add the option to force gzip via an env variable. It will still be doable for the rare cases, but it won't be so invasive. What do you think?

```
os.environ.get('APIFY_PYTHON_CLIENT_FORCE_GZIP', False)
```

vdusek: Sorry, I really don't enjoy overengineering. We just need to be careful here. We are adding new functionality to an open-source tool that has one major release cycle per year, so introducing a feature with the wrong design would be really painful to undo/fix.

Regarding the 'APIFY_PYTHON_CLIENT_FORCE_GZIP' environment variable: I understand that it has a minimal footprint, but it is a hidden global side channel. It does not appear in the type system, it does not compose when someone runs multiple clients in the same process, and it gives us no clear path toward a `quality` setting or additional encodings.

My preference would be to wrap compression in a dependency-injected strategy class and expose a simple `encoding` literal on the facade, with `gzip` as the default.

This also directly addresses my original concern about silent behavior changes: Brotli would only be enabled when the user explicitly requests it, rather than when a transitive dependency happens to install it. This is also the most future-proof option if we would want to add another encoding later (e.g. `zstd`).

Public API:

```python
from apify_client import ApifyClientAsync

# Default: gzip, no extra dependency, and no behavior change.
client = ApifyClientAsync(token='...')

# Explicit opt-in to Brotli. Requires 'apify-client[brotli]'.
client = ApifyClientAsync(token='...', encoding='brotli')
```

The type would be:

```python
encoding: HttpCompressionAlgorithm | HttpCompressor = 'gzip'
```

This would accept either the existing `Literal['gzip', 'brotli']` from `types.py` or an instance of `HttpCompressor` for full control, such as custom quality settings, other potential options, or support for a future algorithm.

This keeps the public API smaller and reduces the number of values we'd need to define and maintain ourselves - minimums, maximums, defaults, and range validation. Those things will be validated by the underlying libraries.

Structurally, I'd mirror the existing `http_clients/` module:

```text
src/apify_client/
├── http_clients/
│   ├── __init__.py
│   ├── _base.py      # HttpClient, HttpClientAsync, HttpResponse
│   └── _impit.py     # ImpitHttpClient, ImpitHttpClientAsync
└── http_compressors/
    ├── __init__.py
    ├── _base.py      # HttpCompressor (ABC)
    ├── _gzip.py      # GzipHttpCompressor
    └── _brotli.py    # BrotliHttpCompressor
```

`_base.py`:

```python
from abc import ABC, abstractmethod


class HttpCompressor(ABC):
    """Strategy for compressing request bodies."""

    content_encoding: str
    """Value sent in the `Content-Encoding` header, for example `gzip` or `br`."""

    @abstractmethod
    def compress(self, data: bytes) -> bytes:
        """Compress a request body."""
```

`_gzip.py`:

```python
import gzip

from apify_client.http_compressors._base import HttpCompressor


class GzipHttpCompressor(HttpCompressor):
    content_encoding = 'gzip'

    def __init__(self, *, quality: int = 6) -> None:
        self._quality = quality

    def compress(self, data: bytes) -> bytes:
        return gzip.compress(bytes(data), compresslevel=self._quality)
```

`_brotli.py` would use a regular top-level import. The `brotli` extra would therefore be required to import this module:

```python
import brotli

from apify_client.http_compressors._base import HttpCompressor


class BrotliHttpCompressor(HttpCompressor):
    content_encoding = 'br'

    def __init__(self, *, quality: int = 6) -> None:
        self._quality = quality

    def compress(self, data: bytes) -> bytes:
        return brotli.compress(bytes(data), quality=self._quality)
```

`__init__.py` would handle the optional extra in the same way as `crawlee-python`, using `try_import` and `install_import_hook`. Importing the package itself would never fail, but accessing `BrotliHttpCompressor` without the extra installed would raise a clear `ImportError`:

```python
from apify_client._utils import install_import_hook, try_import
from apify_client.http_compressors._base import HttpCompressor
from apify_client.http_compressors._gzip import GzipHttpCompressor

# `brotli` is an optional extra. Defer the ImportError until
# BrotliHttpCompressor is actually accessed.
with try_import(__name__, 'BrotliHttpCompressor'):
    from apify_client.http_compressors._brotli import BrotliHttpCompressor

install_import_hook(__name__)

__all__ = ['BrotliHttpCompressor', 'GzipHttpCompressor', 'HttpCompressor']
```

We don't currently have `try_import` here, so just port it to `apify_client/_utils.py` from crawlee-python: https://github.com/apify/crawlee-python/blob/master/src/crawlee/_utils/try_import.py

This gives us:

- Consistency with the codebase: It follows the same structure as `http_clients/` - a base class plus one file per implementation - and uses the same optional-import mechanism as `crawlee-python`.
- Future-proofing: Adding new encoding like `zstd` would mean adding `_zstd.py` and one registry entry. No signature changes or breaking changes would be required. Similar with deprecating encodings.
- Fail-loud behavior instead of silent fallback: Using `encoding='brotli'` without the extra would raise a clear error. The user would either get exactly what they requested or immediately learn why it's unavailable.
- An escape hatch for custom quality settings and implementations: We wouldn't need to add a `quality` parameter to the public facade. Advanced users could inject a configured compressor directly (this is also a pattern we utilize a lot in our tooling repos, mostly in Crawlee):

```python
from apify_client.http_compressors import BrotliHttpCompressor

client = ApifyClientAsync(
    token='...',
    encoding=BrotliHttpCompressor(quality=9),
)
```

In other words, `encoding` would accept either a literal for the convenient path, or an instance of `HttpCompressor` for fine-grained control.

So yeah, it is slightly more complex than an environment variable, but I hope I showed it's the right way to go.

mixalturek: I really like this design, many thanks for the suggestion. Implemented.

vdusek (approved):
I did a few additional improvements. @Pijukatel do you want to re-check it?

docs/02_concepts/13_http_compression.mdx
```diff
@@ -0,0 +1,95 @@
+---
+id: http-compression
+title: HTTP compression
+description: The client compresses every request body automatically using gzip by default, with optional brotli via an explicit opt-in.
+---
+
+The Apify client compresses every request body before sending it to the API. This reduces the amount of data transferred over the network, resulting in faster requests and lower bandwidth usage, especially for large payloads such as Actor inputs, dataset uploads, or key-value store records.
```
szaganek:
```suggestion
The Apify client compresses every request body before sending it to the API. It reduces the amount of data transferred over the network, resulting in faster requests and lower bandwidth usage, especially for large payloads such as Actor inputs, dataset uploads, or key-value store records.
```

docs/02_concepts/13_http_compression.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+## How it works
+
+The client compresses request bodies using the compressor configured via the `compression` parameter (default `'gzip'`). The server supports both gzip and brotli and decompresses the request body transparently.
+
+## Configuration
+
+Pass `compression` to the client constructor to choose the compression algorithm:
```
szaganek:
```suggestion
To choose the compression algorithm, pass `compression` to the client constructor:
```

docs/02_concepts/13_http_compression.mdx
```diff
@@ -0,0 +1,95 @@
...
+| **CPU cost**                  | Moderate, depends on quality            | Low                             |
+| **Availability**              | Requires the `brotli` extra             | Built-in, no extra needed       |
+| **`Content-Encoding` header** | `br`                                    | `gzip`                          |
+| **Quality range**             | `0–11`                                  | `1–9`                           |
+| **Default quality**           | `6`                                     | `9`                             |
+| **Enable via config**         | `compression='brotli'`                  | `compression='gzip'` (default)  |
+| **Best for**                  | Large payloads where bandwidth matters  | Minimal-dependency environments |
+
+:::tip
+For most workloads, the bandwidth savings from brotli outweigh the small CPU overhead. Install the `brotli` extra and pass `compression='brotli'` unless you're running in an environment where installing additional packages isn't feasible.
```
szaganek:
```suggestion
For most workloads, the bandwidth savings from brotli outweigh the CPU costs. Install the `brotli` extra and pass `compression='brotli'` unless you can't install additional packages in your environment.
```
Just simplification that's easier on the eyes

szaganek (approved):

PR #947, opened 2026-07-13
fix: relax required stats fields (compute units, counts) to optional

apify-service-account:
- Updates the auto-generated Pydantic models and TypedDicts based on the proposed OpenAPI specification changes.
- Based on apify-docs PR [#2763](https://github.com/apify/apify-docs/pull/2763).


vdusek (approved):
LGTM. 

I just made the PR title more specific. The previous one ("Relax constraints on generated models based on updated specification") doesn't say much in the changelog (which constraints? which fields?), and we already have another PR with that exact title, so we'd get two identical changelog lines.

Please be more specific in these. And ask Claude if you can't come up with a good title 😄.

PR #948, opened 2026-07-13
feat: support APIFY_API_BASE_URL and APIFY_API_PUBLIC_BASE_URL env vars

Pijukatel:
## Description

The API base URL and public API base URL could only be set through constructor arguments, so pointing the client at a local platform, a staging deployment, or a custom host/port meant threading arguments through every construction site.

Both URLs now resolve as **explicit argument > environment variable > default constant**:

- `APIFY_API_BASE_URL` → `api_url`
- `APIFY_API_PUBLIC_BASE_URL` → `api_public_url`

The two variables are independent (each keeps its own default). Empty-string env values are treated as unset. The resolution is applied identically in the sync `ApifyClient`, the async `ApifyClientAsync`, and the `with_custom_http_client` path, and the automatic `/v2` suffixing is unchanged. **Setting nothing reproduces today's behavior exactly.**

### Context

This is one of three sibling PRs applying the same env-var contract across the Apify clients and CLI (the CLI also adds `APIFY_CONSOLE_BASE_URL`). Separate repos, no overlapping files:

- apify-client-python (this PR): apify/apify-client-python#948
- apify-client-js: apify/apify-client-js#967
- apify-cli: apify/apify-cli#1275

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,23 @@ client = ApifyClient(
 )
 ```
 
+### Point the client at a different API server
+
+```python
+from apify_client import ApifyClient
+
+client = ApifyClient(
+    token='MY-APIFY-TOKEN',
+    api_url='http://localhost:8080',
+)
+```
+
+`api_url` and `api_public_url` also fall back to the `APIFY_API_BASE_URL` and `APIFY_API_PUBLIC_BASE_URL`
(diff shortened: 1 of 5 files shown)


Pijukatel:
Close in favor of smaller CLI update: https://github.com/apify/apify-cli/pull/1276

PR #952, opened 2026-07-16
fix: CamelCase undeclared request fields in add_request and batch_add_requests

vdusek:
`RequestQueueClient.add_request` and `batch_add_requests` validate dict input into the `RequestDraft` model, which declares only `id`, `unique_key`, `url`, and `method`. Every other request field (e.g. `user_data`, `no_retry`) is captured as a Pydantic extra, and extras bypass the `to_camel` alias generator on `model_dump(by_alias=True)`. As a result, snake_case fields shipped to the API verbatim (`user_data` instead of `userData`) and were silently lost. The same dict via `update_request` worked, because `Request` declares the full field set.

The fix adds a `_dump_request_draft()` helper that camelizes top-level extra keys after the aliased dump, used at all four dump sites (sync/async `add_request` and `batch_add_requests`). An extra whose camelCase form is already present in the dump is kept as-is, so it never overwrites another field. Nested content (`user_data` values, headers) is untouched. Regression tests assert the exact JSON body the server receives.

Note: the deeper root cause is that `RequestDraft` under-declares the request fields in the OpenAPI spec, which also leaves the generated `RequestDraftDict` TypedDicts too narrow for typed callers. That is worth a follow-up in apify-docs; this PR fixes the wire behavior on the client side.

### DIFF
--- a/src/apify_client/_resource_clients/request_queue.py
+++ b/src/apify_client/_resource_clients/request_queue.py
@@ -7,6 +7,7 @@ from queue import Queue
 from typing import TYPE_CHECKING, Any, Literal
 
 from more_itertools import constrained_batches
+from pydantic.alias_generators import to_camel
 
 from apify_client._docs import docs_group
 from apify_client._models import (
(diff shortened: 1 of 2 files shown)


vdusek:
closing in favor of https://github.com/apify/apify-docs/pull/2774

PR #953, opened 2026-07-16
fix: Make batch_add_requests split batches by serialized payload size

vdusek:
The 9 MB payload guard in `batch_add_requests` was inert: `constrained_batches` was called without `get_len`, so the default `len()` measured each request dict's key count (~4) instead of its serialized size. Batches were therefore split only by the 25-request count limit, and large requests shipped as one oversized POST that the API rejects with 413, failing the whole call.

The guard now measures each request as its UTF-8 JSON byte length, using the same serialization flags as the HTTP client's request body path. It also passes `strict=False`, which preserves the previous contract for an individually oversized request: it's sent in its own batch and left for the API to judge, instead of raising a client-side `ValueError`. Both the size-based splitting and the oversized-singleton path are covered by new sync/async regression tests.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_resource_clients/request_queue.py
+++ b/src/apify_client/_resource_clients/request_queue.py
@@ -58,19 +58,37 @@ if TYPE_CHECKING:
     from apify_client.types import Timeout
 
 _RQ_MAX_REQUESTS_PER_BATCH = 25
-_MAX_PAYLOAD_SIZE_BYTES = 9 * 1024 * 1024  # 9 MB
-_SAFETY_BUFFER_PERCENT = 0.01 / 100  # 0.01%
+"""Maximum number of requests the API accepts in a single batch call."""
+
+_MAX_PAYLOAD_SIZE_BYTES = 9 * 1024 * 1024
+"""Maximum payload size (9 MB) the API accepts for a single batch call."""
+
+_SAFETY_BUFFER_PERCENT = 0.01 / 100
+"""Safety margin (0.01%) deducted from the maximum payload size when splitting requests into batches."""
 
 
-def _serialize_request(request: dict) -> bytes:
-    """Serialize a request into the JSON bytes it will occupy in the batch request body.
+def _serialize_requests(
+    requests: list[RequestDraft] | list[RequestDraftDict] | list[RequestDraftCamelDict],
+) -> list[bytes]:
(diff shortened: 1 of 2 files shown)


src/apify_client/_resource_clients/request_queue.py
```diff
@@ -61,6 +62,15 @@
 _SAFETY_BUFFER_PERCENT = 0.01 / 100  # 0.01%
 
 
+def _get_payload_size_bytes(request: dict) -> int:
+    """Return the byte size of a request as it will be serialized into the request body.
+
+    Must mirror the JSON serialization in `HttpClientBase._prepare_request_call`, so that measured sizes match the
+    bytes actually sent.
+    """
+    return len(json.dumps(request, ensure_ascii=False, allow_nan=False, default=str).encode('utf-8'))
```
Pijukatel: Is this going to lead to two serializations for each request?
Once to calculate the size and once when sending it in the HTTP client

Pijukatel: Is the max payload size in the API measured before or after potential decompression?

vdusek: You are right, this is not optimal. Fixed.

(btw. JS client does exactly the same)

vdusek: After inspecting the `apify-core`, the limit is being measured after the decompression. Which should mean the decompressed bytes on the API side are exactly the bytes we measure here before compression.

Pijukatel: Could you please create issue or PR for the JS version as well. (If you meant that JS version does the same as double compression :-) )

Pijukatel (approved):
Looks good to me.

Claude had one more minor point:

> Peak memory: requests_as_dicts and serialized_requests are now both alive for the whole call, so peak memory is roughly dicts + full serialized JSON at once. Previously the serialized form existed only per-batch at send time. For crawler-scale inputs (hundreds of thousands of requests) this is a real bump; a cheap mitigation is del requests_as_dicts after serialization. Acceptable tradeoff for exact measurement, but worth a thought.

src/apify_client/_resource_clients/request_queue.py
vdusek: https://github.com/apify/apify-client-js/issues/972

PR #956, opened 2026-07-16
fix: Propagate API token to custom HTTP clients

vdusek:
`ApifyClient.with_custom_http_client(token=...)` (and the async twin) stored the token on the `ApifyClient` instance but never passed it to the injected HTTP client, so no request carried an `Authorization` header and every call failed with 401. The documented custom HTTP client example inherited the bug.

- `with_custom_http_client` now sets `Authorization: Bearer <token>` on the injected client's default headers, unless the client already has an auth header configured (checked case-insensitively).
- `HttpClientBase._prepare_request_call` now merges the client's default headers under the per-request headers, so any custom client using the helper (including a pre-built `ImpitHttpClient` passed as the custom client) actually sends them. For the default client the wire behavior is unchanged, since impit request-level headers replace the identical client-level ones.
- The HTTPX guide examples now merge `self._headers` before delegating, and the `HttpClient` ABC docstring states that implementations must send the default headers with every request.

Regression tests cover the token reaching the wire (custom sync/async clients and a pre-built tokenless `ImpitHttpClient`) and the no-clobber semantics for client-configured auth headers.

### DIFF
--- a/src/apify_client/http_clients/_base.py
+++ b/src/apify_client/http_clients/_base.py
@@ -143,7 +143,7 @@ class HttpClientBase:
         if token is not None:
             default_headers['Authorization'] = f'Bearer {token}'
 
-        self._headers = {**default_headers, **(headers or {})}
+        self._headers = self._merge_headers(default_headers, headers)
 
     def set_default_authorization(self, token: str) -> None:
         """Set the `Authorization` header from the token, unless an authorization header is already configured.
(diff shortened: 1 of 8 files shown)


src/apify_client/http_clients/_base.py
```diff
@@ -207,11 +216,15 @@ def _prepare_request_call(
...
+        """Prepare headers, params, and body for an HTTP request.
+
+        Merges the client's default headers (including authorization) with per-request headers,
+        serializes JSON and compresses the body.
+        """
         if json is not None and data is not None:
             raise ValueError('Cannot pass both "json" and "data" parameters at the same time!')
 
-        headers = dict(headers) if headers else {}
+        headers = {**self._headers, **(headers or {})}
```
Pijukatel: Impit client was working without this, so I guess this is going to duplicate header handling at this level for the default Impit client.

https://github.com/apify/impit/blob/accab5669aa3c3d8e5dc3adc75c847294eabd104/impit/src/impit.rs#L413

vdusek: It won't duplicate the headers because we use `_prepare_request_call`, where we do deduplication. But yeah, it would make sense to update the Impit init, stop passing the default headers, and rely on the upstream behavior.

vdusek:
I'll also update it to handle the fact, that HTTP headers should be case-insensitive (according to RFC 9110). And also add more tests regarding the headers.

Pijukatel (approved):

PR #958, opened 2026-07-17
docs: Fix documentation mismatches with actual client behavior

vdusek:
Fixes 12 documentation issues found by an engineering audit, where docs, examples, or docstrings contradict what the client actually does:

- The custom HTTP client guide example (`HttpxClient`) now raises `ApifyApiError` for error responses instead of returning them raw, and the guide documents this part of the `call` contract (resource clients rely on it, e.g. to translate a 404 into a `None` return value).
- The streaming concepts page no longer claims all three streaming methods yield a raw `impit.Response` — it now describes the actual yielded value per method (`stream_record` yields a `dict` with the response under `value`, `stream` and `stream_record` may yield `None`).
- The pagination concepts page no longer lists `ListOfRequests` among page models exposing `total`/`offset`/`count`; it now explains its cursor-based pagination via `next_cursor`.
- The logging formatter example no longer references `%(status_code)s` (absent on most records, causing logging errors) and no longer attaches a duplicate handler; the page notes which properties are present on every record.
- The upgrading-to-v3 guide cross-links now include the site baseUrl (`/api/client/python/...`), fixing 6 links that 404ed on the published site.
- The conda instruction for the brotli extra installs `brotli-python` (the Python bindings) instead of `brotli` (the C library).
- `RunClient.resurrect` docstrings cite the real `SUCCEEDED` status instead of the nonexistent `FINISHED`.
- `wait_for_finish` docstrings in `RunClient` and `BuildClient` spell the terminal status as `TIMED-OUT` (the real literal) instead of `TIMED_OUT`.
- The quick-start page refers to the `Run` model's `default_dataset_id` attribute instead of the v2-era run dictionary with `defaultDatasetId`.
- The README dataset example passes `fields` as `list[str]` per the signature instead of a comma-separated string.
- The README quick-start examples handle the `Run | None` return of `call()` instead of accessing attributes on a possible `None`.
- The timeouts concepts page states that `no_timeout` is capped at 24 hours by the default client instead of claiming it disables the timeout entirely.

*✍️ Drafted by Claude Code*

### DIFF
--- a/docs/01_introduction/index.mdx
+++ b/docs/01_introduction/index.mdx
@@ -61,6 +61,12 @@ For better request-body compression, opt in to `brotli`, which compresses better
     </TabItem>
 </Tabs>
 
+:::note Extras on conda-forge
+
+Conda doesn't support optional extras. On conda-forge, install the `brotli-python` package directly alongside the client. It provides the same `brotli` module as the PyPI extra.
+
+:::
+
 For details, see [HTTP compression](../02_concepts/13_http_compression.mdx).
 
 ## Quick example
--- a/docs/01_introduction/index.mdx
+++ b/docs/01_introduction/index.mdx
@@ -56,7 +56,7 @@ For better request-body compression, opt in to `brotli`, which compresses better
     </TabItem>
     <TabItem value="conda-forge" label="conda-forge">
         ```bash
-        conda install conda-forge::apify-client conda-forge::brotli
+        conda install conda-forge::apify-client conda-forge::brotli-python
         ```
     </TabItem>
 </Tabs>
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ client = ApifyClient('MY-APIFY-TOKEN')
 run = client.actor('apify/hello-world').call(
     run_input={'message': 'Hello, Apify!'},
 )
+if run is None:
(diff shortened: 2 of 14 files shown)


docs/01_introduction/index.mdx
```diff
@@ -56,7 +56,7 @@ For better request-body compression, opt in to `brotli`, which compresses better
     </TabItem>
     <TabItem value="conda-forge" label="conda-forge">
         ```bash
-        conda install conda-forge::apify-client conda-forge::brotli
+        conda install conda-forge::apify-client conda-forge::brotli-python
```
Pijukatel, line 59: Conda forge does not have any notion of extras dependencies. So I am not sure this will work. It would require some testing. Maybe better to drop all brotli-related advice until tested.

I created a follow-up:
https://github.com/conda-forge/apify-client-feedstock/issues/3

Pijukatel (approved):

docs/01_introduction/index.mdx
vdusek: Yes, it seems Conda-forge doesn't recognize the extras, so installing Brotli directly is the equivalent, isn't it? 

Maybe I can put an additional note there to clarify that.

PR #959, opened 2026-07-20
docs: Remove conda special instructions for brotli

Pijukatel:
It is now included by default: https://github.com/conda-forge/apify-client-feedstock/pull/4/changes

### DIFF
--- a/docs/01_introduction/index.mdx
+++ b/docs/01_introduction/index.mdx
@@ -56,16 +56,10 @@ For better request-body compression, opt in to `brotli`, which compresses better
     </TabItem>
     <TabItem value="conda-forge" label="conda-forge">
         ```bash
-        conda install conda-forge::apify-client conda-forge::brotli-python
+        conda install conda-forge::apify-client
         ```
     </TabItem>
 </Tabs>
-
-:::note Extras on conda-forge
-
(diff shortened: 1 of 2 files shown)


docs/01_introduction/index.mdx
```diff
@@ -56,17 +56,11 @@ For better request-body compression, opt in to `brotli`, which compresses better
...
         ```bash
-        conda install conda-forge::apify-client conda-forge::brotli-python
+        conda install conda-forge::apify-client
         ```
     </TabItem>
 </Tabs>
 
-:::note Extras on conda-forge
-
-Conda doesn't support optional extras. On conda-forge, install the `brotli-python` package directly alongside the client. It provides the same `brotli` module as the PyPI extra.
```
vdusek, line 66: Well, conda doesn't support brotli, that's true, right? So maybe we should just update this note to reflect the current state (brotli is part of the core package if I am not mistaken).

Pijukatel: I think that it is an unnecessary detail. In conda-forge everything is commonly included (in other packages as well), it is common practice since extras are not supported yet. So I think that conda-forge users are expecting this by default.

vdusek (approved):
OK

PR #963, opened 2026-07-20
fix: Normalize query params in dataset create_items_public_url

vdusek:
`DatasetClient.create_items_public_url` (and its async twin) passed `_build_params` output straight to `urlencode`, bypassing the `_parse_params` normalization that the real HTTP request path applies. As a result, boolean and list query params ended up as Python reprs in the signed URL — e.g. `create_items_public_url(clean=True, fields=['title', 'url'])` produced `clean=True&fields=%5B%27title%27%2C+%27url%27%5D` instead of `clean=true&fields=title,url`. Consumers of the shared URL then got unclean/unfiltered items or an API error.

The fix routes the params through `self._http_client._parse_params(...)` before `urlencode`, so the public URL matches exactly what an actual API request would send (bool→`true`/`false`, list→comma-joined, `None` dropped). Added regression tests for both the sync and async clients.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_resource_clients/_resource_client.py
+++ b/src/apify_client/_resource_clients/_resource_client.py
@@ -5,6 +5,7 @@ import time
 from datetime import UTC, datetime, timedelta
 from functools import cached_property
 from typing import TYPE_CHECKING, Any, Literal, get_args
+from urllib.parse import urlencode, urlparse, urlunparse
 
 from apify_client._consts import DEFAULT_WAIT_FOR_FINISH, DEFAULT_WAIT_WHEN_JOB_NOT_EXIST
 from apify_client._docs import docs_group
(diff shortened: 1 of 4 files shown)


vdusek (commented):
It's discutable whether `http_client._parse_params` should remain private. This is its only use case so far, so I'd probably keep it private as it is, but I'm open to discussing it.

Pijukatel (changes requested):
Isn't this just a refactoring? Extracting code used in 6 places into one function. All the new and previous tests are passing even when reverting the code changes in `src`.

src/apify_client/_resource_clients/_resource_client.py
```diff
@@ -115,6 +116,25 @@ def _build_url(
...
+
+        Args:
+            path: Path segment appended to the resource URL (e.g. 'items', 'keys').
+            params: Query parameters to normalize and append.
+
+        Returns:
+            The public URL with a normalized query string.
+        """
+        public_url = urlparse(self._build_url(path, public=True))
+        filtered_params = self._http_client._parse_params(params)  # noqa: SLF001
```
Pijukatel, line 133: We do not need this private import, we can just filter out the None value, or what am I missing?
`{k: v for k, v in params.items() if v is not None}`

Pijukatel: Ok, I see

Pijukatel (approved):
Ok, I mssed some tests

vdusek:
> Ok, I mssed some tests

Also, the PR description mentions it.

PR #964, opened 2026-07-20
fix: Keep pagination iterators advancing past fully-filtered pages

vdusek:
The offset- and cursor-based pagination iterators stopped as soon as a page's `items` list came back empty. But under dataset filters (`clean`, `skip_empty`, `skip_hidden`), the API can *scan* a full chunk (up to 1000 items) while *returning* none — `items=[]` with `x-apify-pagination-count: 1000`. The iterators broke on that page and silently never yielded any item behind it, even though the offset bookkeeping (`max(count, len(items))`) had already advanced past the scanned rows.

Now all four iterators (offset + cursor, sync + async) terminate on the number of items *scanned* rather than *returned*: they keep advancing the offset/cursor across fully-filtered pages and stop only when a page scans nothing. The `limit` and `cursor is None` checks are unchanged. Behavior is identical for every response without a divergent `count` (collection endpoints, `ListOfRequests`).

Added a regression test for each of the four iterators.

Related to https://github.com/apify/apify-core/issues/27324.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_pagination.py
+++ b/src/apify_client/_pagination.py
@@ -41,8 +41,11 @@ def get_items_iterator(
     is used for offset bookkeeping (the Apify API's `count` reflects items scanned, which can exceed items returned when
     filters are applied).
 
-    Iteration stops when a page returns no items or when the user-requested `limit` is reached. The `total` field is
-    intentionally not consulted, because it can change between calls.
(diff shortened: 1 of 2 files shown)


Pijukatel (approved):
Can you please check the JS implementation as well

vdusek:
issue in JS repo https://github.com/apify/apify-client-js/issues/979

PR #965, opened 2026-07-20
fix: Read file-like KVS values before upload and reject unencodable ones

vdusek:
Passing a file-like value (`open(...)`, `io.BytesIO(...)`) to `set_record` — or to `actor.start` / `actor.call` / `actor.validate_input` / `run.metamorph` via `run_input` — crashed with a raw `TypeError: 'BytesIO' object is not an instance of 'Sequence'`. `encode_key_value_store_record_value` set a content type for `io.IOBase` but passed the object through unread, and impit's `content=` only accepts bytes-like bodies.

The encoder now reads file-like values into memory. Detection is duck-typed on a callable `read`, so non-`io.IOBase` wrappers work too. Binary maps to `application/octet-stream`, text to `text/plain`. `read()` is called with no arguments, so the value is consumed from its current position and the object is neither rewound nor closed.

Values it cannot turn into a body now raise `TypeError` instead of reaching the transport:

- async file-likes (`aiofiles`, Starlette `UploadFile`) — `read()` returns a coroutine
- `read()` returning neither bytes nor `str`, e.g. `None` from a non-blocking raw stream
- a non-bytes-like value with a non-JSON content type, e.g. `set_record('k', {'a': 1}, content_type='image/png')`

Note: impit exposes no streaming `content=` API, so uploads are fully buffered rather than streamed.

Tests: `set_record` over `BytesIO` / `StringIO` / a duck-typed reader, sync and async, over both compression algorithms, plus encoder coverage of the three rejection paths.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_resource_clients/key_value_store.py
+++ b/src/apify_client/_resource_clients/key_value_store.py
@@ -802,8 +802,7 @@ class KeyValueStoreClientAsync(ResourceClientAsync):
             content_type: The content type of the saved value.
             timeout: Timeout for the API HTTP request.
         """
-        # Encoding reads file-like values and may serialize large payloads, which is blocking; offload it to a
-        # worker thread so it does not stall the event loop (mirrors the transport's own body-prep offload).
+        # Encoding may read a file or serialize a large payload (blocking), so run it off the event loop.
         value, content_type = await asyncio.to_thread(
             encode_key_value_store_record_value, value, content_type=content_type
         )
--- a/src/apify_client/_resource_clients/key_value_store.py
+++ b/src/apify_client/_resource_clients/key_value_store.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import re
 from contextlib import asynccontextmanager, contextmanager
 from http import HTTPStatus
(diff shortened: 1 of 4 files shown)


src/apify_client/_resource_clients/key_value_store.py
```diff
@@ -801,7 +802,10 @@ async def set_record(
             content_type: The content type of the saved value.
             timeout: Timeout for the API HTTP request.
         """
-        value, content_type = encode_key_value_store_record_value(value, content_type=content_type)
+        # Encoding may read a file or serialize a large payload (blocking), so run it off the event loop.
+        value, content_type = await asyncio.to_thread(
```
Pijukatel: `asyncio.to_thread` is useful only to the branch that can do IO. It just adds overhead in other cases, or? So it improves the less frequent case at the cost of the main use case

vdusek: Well, AFAIK the criterion is more whether the callee releases the GIL. If it does, the worker thread can work on the offloaded task, while the main thread, which runs the event loop, can run other asyncio tasks.

And here is the problem: `json.dumps`, which is called in `encode_key_value_store_record_value`, never releases the GIL. So yeah, let's remove it.

Thanks.

Mantisus (approved):
LGTM

B4nan (approved):

PR #971, opened 2026-07-21
test: deflake test_actor_create_update_delete against retried create conflict

vdusek:
The integration test `test_actor_create_update_delete` intermittently failed with `ConflictError: Some other Actor already has this name`, e.g. in [this CI run](https://github.com/apify/apify-client-python/actions/runs/29811521395/job/88573317358) (on an unrelated Renovate PR).

Root cause: the HTTP client retries every request on transient 5xx and network errors (`max_retries=8`). A `POST /v2/actors` create can commit server-side on one attempt yet still be retried; the retry then hits a 409 on the unique name it just created. The failing traceback shows `attempt=2`, and the 61^8 random resource name rules out cross-worker or leftover collisions.

Create calls in the integration tests now go through conflict-recovering helpers in `tests/_utils.py`. On the API error that signals a lost name or version number, the helper fetches what the first attempt committed and the test continues. Anything else propagates, and so does a matching error whose resource cannot be found afterwards - a genuine create failure still fails the test rather than being masked.

This covers every create of a resource with a uniqueness constraint: Actors (`actor-name-not-unique`), tasks (`actor-task-name-not-unique`), schedules (`schedule-name-not-unique`), Actor versions (`version-already-exists`) and env vars (`env-var-already-exists`). Webhooks have no name uniqueness, and the dataset/KVS/queue creates use `get_or_create`, so neither needs a helper.

The recovery branches only run when a retried create already committed, which never happens in a normal integration run. `tests/unit/test_create_helpers.py` therefore drives each helper against a mocked API, so a wrong error type or lookup route is caught there instead of resurfacing as the flake the helpers exist to prevent.

This is a test-side fix. The underlying client behaviour - retrying non-idempotent requests - is tracked in #990; besides these spurious conflicts it can also silently duplicate run starts and webhook creations for callers.

*✍️ Drafted by Claude Code*


Pijukatel (approved):

vdusek:
converting to draft as I'm not sure whether these test utils are worth it, let me @Pijukatel know WDYT

vdusek:
This is probably something we should not handle on the API client side. I opened an issue instead https://github.com/apify/apify-client-python/issues/990. It might be an API-side bug that should be fixed there.

PR #936, opened 2026-07-09
fix: Relax constraints on generated models based on updated specification

apify-service-account:
- Updates the auto-generated Pydantic models and TypedDicts based on the proposed OpenAPI specification changes.
- Based on apify-docs PR [#2754](https://github.com/apify/apify-docs/pull/2754).

PR #923, opened 2026-07-07
docs: fix dataset schema link in generated Dataset model

apify-service-account:
- Updates the auto-generated Pydantic models and TypedDicts based on the proposed OpenAPI specification changes.
- Based on apify-docs PR [#2747](https://github.com/apify/apify-docs/pull/2747).

PR #979, opened 2026-07-28
fix: Mark generated model fields that the API can omit as optional

apify-service-account:
Regenerates the auto-generated Pydantic models and TypedDicts from the latest [published OpenAPI specification](https://docs.apify.com/api/openapi.json).

The spec change mostly relaxes fields that were declared required but can be absent in real API responses, so deserializing such responses no longer raises a `ValidationError`:

- `BuildStats.compute_units` and `RunStats.compute_units` are now optional.
- `RunStats.restart_count` and `RunStats.resurrect_count` are now optional.
- `KeyValueStoreStats.read_count`, `write_count`, `delete_count` and `list_count` are now optional.
- `WebhookStats.total_dispatches` is now optional.
- `Webhook.request_url` is now optional - it is omitted or `null` for hook actions other than the conventional HTTP case (e.g. Slack or email notifications). Documented on both `Webhook` and `WebhookDispatchWebhookSummary`.
- `StoreListActor.current_pricing_info` is now optional.
- `UserPrivateInfo.proxy` is now optional.
- `RunOptions.max_items` now allows `0` (constraint relaxed from `ge=1` to `ge=0`).
- New error code literal: `cannot-monetize-without-payout-billing-info`.
- Docs: `TaggedBuildInfo.build_number` notes it can be `null` for legacy builds; the `Dataset.schema` docs link now points to `https://docs.apify.com/storage/dataset-schema`.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_literals.py
+++ b/src/apify_client/_literals.py
@@ -455,7 +455,6 @@ RunOrigin = Literal[
     'CLI',
     'CI',
     'STANDBY',
-    'MCP',
 ]
 
 
--- a/src/apify_client/_literals.py
+++ b/src/apify_client/_literals.py
@@ -88,6 +88,7 @@ ErrorType = Literal[
     'cannot-metamorph-to-pay-per-result-actor',
     'cannot-modify-actor-pricing-too-frequently',
     'cannot-modify-actor-pricing-with-immediate-effect',
+    'cannot-monetize-without-payout-billing-info',
     'cannot-override-paid-actor-trial',
     'cannot-permanently-delete-subscription',
     'cannot-publish-actor',
@@ -455,6 +456,7 @@ RunOrigin = Literal[
     'CLI',
     'CI',
     'STANDBY',
+    'MCP',
 ]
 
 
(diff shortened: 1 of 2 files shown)


vdusek (approved):

vdusek:
Closing as superseded - this PR no longer carries any changes.

Everything it contained has already landed on `master` through separate PRs: #923, #936, #941, #947, #960, #974. After merging `master` into this branch and resolving the `_literals.py` conflict, the resulting tree is byte-identical to `master`.

The conflict was a symptom of a workflow bug rather than a real spec change. The manual regeneration branch (`update-models-manual`) is reused indefinitely and never rebased onto `master`, so the second run checked out the 20-day-old branch tip and regenerated the models with the codegen tooling from that stale commit. That tooling predates #941, so `_literals.py` came out in the old closed `Literal[...]` form - merging this PR would have silently reverted the enum relaxation from #941.

The underlying issues in `manual_regenerate_models.yaml` will be addressed separately.

*✍️ Drafted by Claude Code*

PR #983, opened 2026-07-29
ci: Add nightly model regeneration from the published OpenAPI spec

vdusek:
Replaces the apify-docs-triggered model regeneration with a nightly job, and records which published OpenAPI specification the generated models came from.

## What changes

- **`on_schedule_regenerate_models.yaml`** - nightly at 02:00 UTC plus manual dispatch. Regenerates on master and opens a pull request when the models change. `manual_regenerate_models.yaml` is deleted.
- **`scripts/openapi_spec.py`** - downloads and validates the published specification into git-ignored `tmp/` (both codegen passes read that one copy), then records its `info.version` in `pyproject.toml` under `[tool.apify.openapi-spec]`. The specification itself is not committed.
- **`_models.py`** loses the `Plan.available_proxy_groups` docstring - accumulated codegen drift from the `datamodel-code-generator` bumps (#968, #975), not a specification change.

## Why

The old workflow was dispatched by apify-docs and checked out a permanent, never-rebased branch *before* regenerating, so it generated with that branch's stale tooling. In #979 that produced `_literals.py` in the pre-#941 closed `Literal[...]` form; merging it would have silently reverted the enum relaxation. It also opened with a `TODO` title that blocked `pr-title-check` - a design that needs human action to become mergeable, which is what let #979 rot for 20 days while the same spec changes were landed by hand (#923, #936, #947, #960, #974).

The nightly job always generates on master and rebuilds its `ci/regenerate-models` branch from master instead of appending, so the diff is always "current spec vs current master" and can't resurrect a stale generated file. It opens with a mergeable `chore:` title; reviewers retitle to `fix:`/`feat:` when the diff is user-facing.

## Notes

- **Merge apify/apify-docs#2835 first** - it removes the dispatch that still calls the workflow deleted here.
- **Needs a `SLACK_WEBHOOK_URL` repository secret** for the failure alert.
- The recorded version is a coarse marker, not a content identity: apify-docs bumps `components/version.yaml` in a follow-up `[skip ci]` commit, so a deploy can publish new content under the old stamp. A moved value proves the specification changed; an unchanged one proves nothing.

*✍️ Drafted by Claude Code*


janbuchar (approved):
Looks sound at a glance, let's battle test it

Issue #972, opened 2026-07-21
Support streaming request bodies to avoid buffering large uploads in memory


vdusek:
File-like values passed to `KeyValueStoreClient.set_record` (and Actor run inputs) are fully read into memory by `value.read()` in `encode_key_value_store_record_value`, and `_prepare_request_call` then compresses the whole buffer into a second full copy. Uploading a large record (e.g. multi-GB) from a file handle needs ~2x its size resident in RAM.

## Root cause

There is no chunked/streaming path anywhere in the stack:

- `HttpClient.call` / `HttpClientAsync.call` type the body as `data: str | bytes | bytearray | None` — no iterator or file object.
- `HttpClientBase._prepare_request_call` reads the whole `data` and compresses it in one shot: `self._http_compressor.compress(data)` runs over the full buffer, regardless of the configured `HttpCompressor` (applies to both gzip and brotli).
- The default transport (`ImpitHttpClient`) only accepts a bytes-like `content=` body.

## Notes on a fix

Full streaming requires changes across the whole layer, not just the transport:

1. Accept a streaming body (bytes iterator / async iterator / file object) through the `HttpClient` abstraction.
2. Make compression incremental or skippable, so the body is not re-buffered. This has to hold for every `HttpCompressor` implementation (gzip and brotli), not a single one.
3. Use a transport that can stream the body. `httpx` supports this today — `content=` accepts a sync/async bytes generator and streams it chunk-by-chunk (multipart uploads via `files=` stream by default). It is unclear whether `impit` exposes an equivalent; that needs checking, and `impit` was chosen for its browser-like TLS fingerprints, so swapping transports has trade-offs.

Follow-up to commit bc092b8 (file-like read fix), where this was noted as out of scope.

*✍️ Drafted by Claude Code*

vdusek:
Basically, we should:
- Skip the compression for the streams.
- Provide the stream to the underlying HTTP client.

If I am not mistaken, this is done in the `apify-client-js` using `axios` as the HTTP client.

@barjin Does `impit` support streaming request bodies? According to Claude, it doesn't seem to, but I would prefer an answer from you 😄.

barjin:
That's right, `impit` currently only streams responses in Python. Sounds like a useful addition, though 👍  I added an [issue here](https://github.com/apify/impit/issues/504).

Issue #934, opened 2026-07-09
Skip request-body compression for small payloads


vdusek:
The Python API client compresses every request body regardless of size (see #927). For small payloads the compression overhead isn't worth it. 

The JS client skips compression when the body is below **1024 B**. We should align the Python client with that behavior and only compress bodies at or above the threshold. 

Context: https://github.com/apify/apify-client-python/pull/927#issuecomment-4916516316

Pijukatel:
Maybe it is also a good time to think (make some benchmarks) whether it should be done the other way around and turn it on in JS for all payloads as well.

PR #987, opened 2026-08-03
perf: Skip request-body compression for already-compressed content types

vdusek:
### What

`_prepare_request_call` now checks the request's `Content-Type` before compressing the body and skips compression for media types that already carry their own compression:

- any `image/*`, `audio/*`, or `video/*` type,
- archives (`application/zip`, `application/gzip`, `application/x-7z-compressed`, rar, bzip2, xz, zstd, jar),
- web fonts (`font/woff`, `font/woff2`).

Types with a `+json`/`+xml` structured suffix stay compressible, so `image/svg+xml` is not caught by the `image/` prefix. The media type is normalized before matching (parameters stripped, lowercased, trimmed). When compression is skipped, any caller-supplied `Content-Encoding` header is dropped as well — the body goes out verbatim, so the header would otherwise misdescribe it.

### Why

The client compressed every request body unconditionally. For already-compressed payloads that burns CPU and holds a second full copy of the body in memory, while typically producing output slightly *larger* than the input. Key-value store records — screenshots, video, archives — are exactly this case.

End-to-end `set_record` with a 200 MB incompressible payload:

| Content type | Peak RSS | Wall / CPU |
|---|---|---|
| `application/octet-stream` (unchanged) | 604 MB - 3.02x | 7.8 s / 6.9 s user |
| `image/png` (new path) | 403 MB - 2.02x | 1.05 s / 0.40 s user |

`application/octet-stream` is deliberately not on the list: it is the catch-all for unknown binary, which may well be uncompressed data. It is also the encoder's default when no `content_type` is passed, so the optimization only applies when the caller sets an accurate content type — documented in the compression concept page.

The remaining 2x comes from handing a `bytes` body to impit's `content=`, tracked separately in #972.

### Notes

This overlaps with #934 (skip compression for small payloads) — both add a condition to the same block in `_prepare_request_call`. Whichever lands second needs the two conditions combined.

*✍️ Drafted by Claude Code*


src/apify_client/_consts.py
```diff
@@ -34,3 +34,54 @@
...
+        'application/x-zip-compressed',
+        'application/zip',
+        'application/zstd',
+        'font/woff',
+        'font/woff2',
+    }
+)
+"""Exact media types whose payloads carry their own compression."""
+
+COMPRESSIBLE_MEDIA_TYPES = frozenset(
```
Mantisus, line 66: I would also add `audio/L24` and `audio/midi`

src/apify_client/http_clients/_base.py
```diff
@@ -232,29 +233,41 @@ def _prepare_request_call(
...
+            content_type = next((value for key, value in headers.items() if key.lower() == 'content-type'), None)
+            if is_compressible_content_type(content_type):
+                data = self._http_compressor.compress(data)
+                headers = self._merge_headers(headers, {'Content-Encoding': self._http_compressor.content_encoding})
+                compressed = True
+
+        # Anything left uncompressed goes out as-is - a file-like body included - so a caller-supplied encoding
+        # would misdescribe it.
+        if data is not None and not compressed:
+            headers = {key: value for key, value in headers.items() if key.lower() != 'content-encoding'}
```
Mantisus, line 270: Just out of curiosity. Am I correct in understanding that we don't allow a user to compress the data themselves and send a `Content-Encoding` header?

Mantisus (approved):
LGTM. Only a small suggestion.

src/apify_client/http_clients/_base.py
vdusek: Hmm, good point, let's fix it in a separate PR

src/apify_client/_consts.py
vdusek: added

B4nan (approved):

PR #995, opened 2026-08-03
chore: Regenerate models from the published OpenAPI spec

apify-service-account:
- Regenerates the Pydantic models, TypedDicts, and literal aliases from the [published OpenAPI specification](https://docs.apify.com/api/openapi.json), and records its version in `pyproject.toml`.
- Specification version: `v2-2026-07-28T083939Z` - **unchanged**, which doesn't rule out a specification change: the published stamp can lag its content by a deploy. Only the diff tells.

> [!IMPORTANT]
> Retitle this pull request to `fix:` or `feat:` when the diff is user-facing, so that it lands in the changelog and triggers a release - `chore:` does neither.

> Generated by the [Regenerate models](https://github.com/apify/apify-client-python/actions/workflows/on_schedule_regenerate_models.yaml) workflow.

### DIFF
--- a/src/apify_client/_models.py
+++ b/src/apify_client/_models.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from enum import Enum
 from typing import Annotated, Any, Literal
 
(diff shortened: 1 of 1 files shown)


vdusek:
new pipeline seems to be working

PR #997, opened 2026-08-03
fix: Respect caller-supplied Content-Encoding for pre-compressed request bodies

### LINKED ISSUE #996: Support caller-supplied `Content-Encoding` for pre-compressed request bodies
Follow-up from a review comment on #987: https://github.com/apify/apify-client-python/pull/987#discussion_r3704749196

## Problem

There is no way for a caller to hand the client an already-encoded request body.

- Before #987, `_prepare_request_call` compressed every bytes-like body and merged its own `Content-Encoding` last, so a caller-supplied value was silently overwritten.
- Since #987, a caller-supplied `Content-Encoding` is dropped whenever compression is skipped, so a pre-compressed body goes out labeled as unencoded.

The API supports this and documents it. [`PUT key-value-store record`](https://docs.apify.com/api/v2/key-value-store-record-put) documents a `Content-Encoding` request header accepting `gzip`, `br`, `deflate`, and `identity`: "To save bandwidth, storage, and speed up your upload, send the request payload compressed with Gzip compression and add the `Content-Encoding: gzip` header. It is possible to set up another compression type with `Content-Encoding` request header." And per the [key-value store docs](https://docs.apify.com/platform/storage/key-value-store), "records are stored exactly as you upload them", with the encoding taken from that header — so on that endpoint the header is record metadata, not just a transport hint, and dropping it stores compressed bytes under an unencoded label.

Honoring the header also unlocks encodings the client ships no compressor for (`deflate`, `zstd`, ...) without having to add one.

## How the JS client solves it

`apify-client-js` treats a caller-supplied `content-encoding` as "hands off" — [`src/interceptors.ts#L81-L82`](https://github.com/apify/apify-client-js/blob/fd619c056e64743843ece519fe56ae87f0c1e687/src/interceptors.ts#L81-L82):

```ts
async function maybeCompressRequest(config: ApifyRequestConfig): Promise<ApifyRequestConfig> {
    if (config.headers?.['content-encoding']) return config;

    const maybeCompressed = await maybeCompressValue(config.data);
    ...
}
```

No opt-out parameter and no dedicated compressor — the header alone is the signal. (`maybeCompressValue` also skips bodies below `MIN_COMPRESS_BYTES = 1024`, which is #934 on the Python side.)

## Proposal

1. **Honor a caller-supplied `Content-Encoding` in `_prepare_request_call`**, matching JS. Precedence: caller `Content-Encoding` -> content-type heuristic (`is_compressible_content_type`) -> compress. The header is forwarded verbatim, so any encoding the backend accepts works, and `identity` becomes a per-request opt-out for free.

    ```python
    caller_encoding = next((v for k, v in headers.items() if k.lower() == 'content-encoding'), None)
    if caller_encoding is None and is_compressible_content_type(content_type):
        data = self._http_compressor.compress(data)
        headers = self._merge_headers(headers, {'Content-Encoding': self._http_compressor.content_encoding})
    ```

    Explicitly out of scope: stacking the client's encoding on top of the caller's (`Content-Encoding: gzip, br`) — the backend does not document the list form.

    This reverses the drop-the-header behavior added in #987 and flips `test_prepare_request_call_replaces_caller_content_encoding` plus the two `..._drops_caller_content_encoding_...` tests.

2. **Expose `content_encoding` on `KeyValueStoreClient.set_record`** (sync and async). Step 1 on its own is barely reachable: no public resource method takes per-request headers, so the only entry points are the client-wide `ApifyClient(headers=...)`, which would apply the header to every request, and the private `_http_client.call`. `set_record` is where the API documents the header, so that is where it belongs.

Rejected: a per-request `compress=False` keyword on `HttpClient.call`. `HttpClient`/`HttpClientAsync` are a public ABC, and once resource clients pass the new keyword, existing custom implementations break. Headers already flow through, so step 1 covers the same need without touching that contract.

Docs: add a pre-compressed bodies section to the HTTP compression concept page.

*✍️ Drafted by Claude Code*

vdusek:
A caller-supplied `Content-Encoding` now marks the request body as already encoded, so the client forwards it untouched instead of dropping the header (since #987) or overwriting it (before #987).

## How it works

- Precedence in `_prepare_request_call`: caller `Content-Encoding` -> content-type heuristic -> size threshold -> compress. The header alone is the "hands off" signal, as in `apify-client-js`, so no opt-out parameter lands on the public `HttpClient` ABC. Lookups go through a new case-insensitive `_get_header`.
- The header is forwarded verbatim, so encodings the client ships no compressor for (`deflate`) work too, and `identity` becomes a per-request opt-out.
- `KeyValueStoreClient.set_record` gains a `content_encoding` argument (sync and async). Without it the behavior is barely reachable, as no public resource method took per-request headers.

## Notes

- Not breaking: #987 and #988 are both still in the unreleased 3.1.2 section, so the header-dropping behavior never shipped.
- Three tests from #987 flip. Two collapse into one parametrized `test_prepare_request_call_keeps_caller_content_encoding` covering their cases plus `identity` and `deflate`.
- A `Content-Encoding` passed to the client constructor counts as caller-supplied on every request, as in JS, where axios config headers include the client defaults.
- The compression guide drops a claim that file-like values are streamed and never compressed. Since #965 they're read into memory and compressed like any other body.
- No integration test: the round trip needs a real token, so the API-side premise (a `PUT` record honors the header and serves the record with it) rests on the API docs. The unit tests cover the client side in full.

Closes #996

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_resource_clients/key_value_store.py
+++ b/src/apify_client/_resource_clients/key_value_store.py
@@ -360,6 +360,7 @@ class KeyValueStoreClient(ResourceClient):
         value: Any,
         *,
         content_type: str | None = None,
+        content_encoding: str | None = None,
         timeout: Timeout = 'long',
     ) -> None:
         """Set a value to the given record in the key-value store.
@@ -370,11 +371,17 @@ class KeyValueStoreClient(ResourceClient):
             key: The key of the record to save the value to.
             value: The value to save into the record.
             content_type: The content type of the saved value.
+            content_encoding: The encoding already applied to `value`, sent as the `Content-Encoding` header. Pass it
+                to upload a pre-compressed value - the client then forwards the bytes as they are instead of
+                compressing them itself. The API accepts `gzip`, `br`, `deflate`, and `identity`, and stores the
+                record exactly as uploaded, so this also becomes the encoding the record is served with.
             timeout: Timeout for the API HTTP request.
         """
         value, content_type = encode_key_value_store_record_value(value, content_type=content_type)
 
         headers = {'content-type': content_type}
+        if content_encoding is not None:
+            headers['content-encoding'] = content_encoding
 
         self._http_client.call(
             url=self._build_url(f'records/{key}'),
@@ -776,6 +783,7 @@ class KeyValueStoreClientAsync(ResourceClientAsync):
         value: Any,
         *,
         content_type: str | None = None,
+        content_encoding: str | None = None,
         timeout: Timeout = 'long',
     ) -> None:
         """Set a value to the given record in the key-value store.
@@ -786,11 +794,17 @@ class KeyValueStoreClientAsync(ResourceClientAsync):
             key: The key of the record to save the value to.
             value: The value to save into the record.
             content_type: The content type of the saved value.
+            content_encoding: The encoding already applied to `value`, sent as the `Content-Encoding` header. Pass it
+                to upload a pre-compressed value - the client then forwards the bytes as they are instead of
+                compressing them itself. The API accepts `gzip`, `br`, `deflate`, and `identity`, and stores the
+                record exactly as uploaded, so this also becomes the encoding the record is served with.
             timeout: Timeout for the API HTTP request.
         """
         value, content_type = encode_key_value_store_record_value(value, content_type=content_type)
 
         headers = {'content-type': content_type}
+        if content_encoding is not None:
+            headers['content-encoding'] = content_encoding
 
         await self._http_client.call(
             url=self._build_url(f'records/{key}'),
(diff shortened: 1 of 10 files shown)


src/apify_client/_resource_clients/key_value_store.py
```diff
@@ -786,11 +794,17 @@ async def set_record(
             key: The key of the record to save the value to.
             value: The value to save into the record.
             content_type: The content type of the saved value.
+            content_encoding: The encoding already applied to `value`, sent as the `Content-Encoding` header. Pass it
+                to upload a pre-compressed value - the client then forwards the bytes as they are instead of
+                compressing them itself. The API accepts `gzip`, `br`, `deflate`, and `identity`, and stores the
+                record exactly as uploaded, so this also becomes the encoding the record is served with.
             timeout: Timeout for the API HTTP request.
         """
```
Mantisus, line 810: Could we add a guard that rejects a non-byte `value` (and non-file?) when `content_encoding` declares a compression?

Mantisus (approved):
One nit, otherwise LGTM

B4nan (approved):

PR #1004, opened 2026-08-11
feat: Add HTTPX-based HTTP client

vdusek:
Adds `HttpxHttpClient` / `HttpxHttpClientAsync` as built-in alternatives to the default Impit transport, behind an optional extra: `pip install "apify-client[httpx]"`. Impit stays the default.

The transport is [HTTPX2](https://github.com/pydantic/httpx2), Pydantic's maintained continuation of HTTPX (stable HTTPX has been parked at 0.28.1 since Dec 2024). The module imports it as `httpx`, so the client keeps the plain `Httpx*` naming. It verifies TLS via the OS trust store (`truststore`) rather than `certifi`.

The adapters plug into the shared request pipeline from #1022 and implement only the transport hooks (`send_request`, `is_retryable_transport_error`, `is_timeout_error`, `close()` / `aclose()`), applying the same retry policy as Impit: every error from the library's own hierarchy is transient except the permanently-failing types. A response event hook keeps the shared cookie jar from leaking server cookies into later API requests. Unit tests run the built-in client matrix (Impit + HTTPX) through the parametrized fixtures.

Note that HTTPX applies timeouts per socket operation, not to the request as a whole, so a slowly-arriving body can outlast the requested timeout. Impit enforces a whole-request deadline. Both class docstrings record this.

Stack: #1022 (merged) → **#1004 (this)** → #1013 (docs — its `apify-client[httpx]` install lines now match; only the `conda-forge::httpx` line still points at the stale package instead of `httpx2`).

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/http_clients/__init__.py
+++ b/src/apify_client/http_clients/__init__.py
@@ -5,23 +5,23 @@ from apify_client.http_clients._impit import ImpitHttpClient, ImpitHttpClientAsy
 
 _install_import_hook(__name__)
 
-# `httpx2` is an optional extra, so it's wrapped in try_import. Accessing the HTTPX2 clients
+# `httpx2` is an optional extra, so it's wrapped in try_import. Accessing the HTTPX clients
 # without the extra installed raises a clear ImportError instead of failing at package import time.
 with _try_import(
     __name__,
-    'Httpx2HttpClient',
-    'Httpx2HttpClientAsync',
+    'HttpxHttpClient',
+    'HttpxHttpClientAsync',
     dependency_name='httpx2',
-) as _httpx2_import:
-    from apify_client.http_clients._httpx2 import Httpx2HttpClient, Httpx2HttpClientAsync
+) as _httpx_import:
+    from apify_client.http_clients._httpx import HttpxHttpClient, HttpxHttpClientAsync
 
-if _httpx2_import.available:
+if _httpx_import.available:
     __all__ = [
         'HttpClient',
         'HttpClientAsync',
         'HttpResponse',
-        'Httpx2HttpClient',
-        'Httpx2HttpClientAsync',
+        'HttpxHttpClient',
+        'HttpxHttpClientAsync',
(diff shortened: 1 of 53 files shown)


src/apify_client/http_clients/__init__.py
```diff
@@ -1,10 +1,36 @@
...
+# import time.
+with _try_import(
+    __name__,
+    'HttpxHttpClient',
+    'HttpxHttpClientAsync',
+    dependency_name='httpx2',
+) as _httpx_import:
+    from apify_client.http_clients._httpx import HttpxHttpClient, HttpxHttpClientAsync
+
+if _httpx_import.available:
```
Pijukatel, line 27: Maybe share the default and only extend. Less duplication and it is clear from first look what is the optional

```
__all__ = ...

if _httpx_import.available:
    __all__.extend(...)
```

tests/integration/conftest.py
```diff
@@ -104,23 +106,17 @@ def test_kvs_of_another_user(api_token_2: str) -> Generator[KvsFixture]:
...
-# ============================================================================
-# Function-scoped fixtures (created for each test)
-# ============================================================================
-
-
-@pytest.fixture
-def apify_client(api_token: str) -> ApifyClient:
-    """Sync Apify client instance."""
-    api_url = os.getenv(API_URL_ENV_VAR) or DEFAULT_API_URL
-    return ApifyClient(api_token, api_url=api_url)
```
Pijukatel, line 116: Small concern. The default most common constructor will no longer be well covered by the integration tests.

`ApifyClient(api_token, api_url=api_url)`
`ApifyClientAsync(api_token, api_url=api_url)`

Pijukatel (approved):
Two minor comments

tests/integration/conftest.py
vdusek: I added a new test (in case I understand your concern correctly)

Pijukatel (approved):

PR #1005, opened 2026-08-12
docs: Document URL field normalization in typed models

### LINKED ISSUE #999: URL fields get an extra trailing slash in v3, changing the value the API sent
Title: URL fields get an extra trailing slash in v3, changing the value the API sent

### Issue description

In v3, URL fields are typed as `pydantic.AnyUrl`. Pydantic appends `/` to a URL
that has no path, so the string the client gives back is not the string the API
sent.

Fetching the same actor run two ways shows the difference. The raw API response
returns `containerUrl` as `https://<id>.runs.apify.net`, with no trailing slash.
Reading the same field through `client.run(...).get().container_url` in v3 gives
`https://<id>.runs.apify.net/`, with one added.

The change happens during validation, not serialization, so every access path
carries it - attribute access, `model_dump()` in either mode, and
`model_dump_json()`.

### This is a change from v2

v2 returned plain `dict`s (`ActorClient.call()` and `RunClient.get()` are
annotated `dict | None`) and did not depend on pydantic at all, so the value
passed through untouched. Code that upgrades to v3 silently starts seeing a
different string for the same run.

### Scope

18 `AnyUrl`-typed fields across `apify_client._models` are affected, including
`Run.container_url`, `Task.standby_url`, `Dataset.console_url`,
`KeyValueStore.records_public_url`, `RequestQueue.console_url`, and
`Webhook.request_url`.

### Why this is a problem

I realize `https://example.com` and `https://example.com/` are equivalent per
RFC 3986 section 6.2.3, so this may look harmless. Two reasons why it still matters:

1. **It changes silently across a major version.** Any code that stored a URL
   under v2 and compares it against a freshly fetched one under v3 breaks on
   upgrade, with nothing in the changelog to explain it.
2. **A client shouldn't alter values the server sent.** Even where the two forms
   are equivalent as URLs, they are not equal as strings, and callers reasonably
   expect the field to match the API response.

Anywhere the value is used as a base to build longer URLs, the added slash also
produces a doubled separator (`...net//path`), which is not equivalent.

We hit this in the Apify Dify plugin and work around it like so:

```python
parts = urlsplit(container_url)
if parts.path == "/":
    payload["containerUrl"] = urlunsplit(parts._replace(path=""))

vdusek:
URL fields on the returned models are `pydantic.AnyUrl`, which normalizes the value during validation, so the string read back can differ from the one the API sent. The trailing slash is the most visible case; normalization also lowercases the host, drops default ports, punycodes internationalized hosts, and percent-encodes unsafe characters. The behavior is intended, it just wasn't documented.

- `docs/02_concepts/12_typed_models.mdx` gets a "URL fields" section covering the normalization cases, comparing URLs as `AnyUrl` rather than raw strings, and building longer URLs with `urljoin`.
- `docs/04_upgrading/upgrading_to_v3.mdx` gets a "URL fields are normalized" section framing it as the v2 to v3 change, with the affected field names.

Closes #999

*✍️ Drafted by Claude Code*

### DIFF
--- a/docs/02_concepts/12_typed_models.mdx
+++ b/docs/02_concepts/12_typed_models.mdx
@@ -36,6 +36,50 @@ Every method that returns a structured payload returns a Pydantic model. Fields 
 
 Date strings are automatically parsed into timezone-aware `datetime.datetime` objects, enums into `Literal` aliases, and nested objects into their own typed models, so you can compose attribute access without manual conversion.
 
+## URL fields
+
+Fields holding a URL, such as `Run.container_url` or `Dataset.console_url`, are typed as [`AnyUrl`](https://docs.pydantic.dev/latest/api/networks/#pydantic.networks.AnyUrl). Their values are validated and normalized, so the string you read back can differ from the one the API sent. Both forms denote the same URL under [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.3), they're just different strings.
+
+```python
+from pydantic import AnyUrl
+
+# An empty path becomes '/'.
+str(AnyUrl('https://abc123.runs.apify.net'))  # 'https://abc123.runs.apify.net/'
+
+# The host is lowercased.
+str(AnyUrl('https://EXAMPLE.com/Path'))  # 'https://example.com/Path'
+
+# A default port is dropped.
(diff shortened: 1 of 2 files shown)


Pijukatel (approved):
Nice with all the examples

PR #1006, opened 2026-08-12
fix: Unify the request pipeline and retry policy across HTTP clients

vdusek:
Moves `call`, the retry loop, and the per-attempt request handling from `ImpitHttpClient` into `HttpClient` and `HttpClientAsync`, leaving Impit as a thin adapter over the transport hooks: `send_request`, `is_retryable_transport_error`, `is_timeout_error`, and `close()` / `aclose()` plus context managers. A custom client that overrides `call` keeps working unchanged and opts out of the shared pipeline.

The same pass fixes:

- Transport errors a retry cannot fix (`UnsupportedProtocol`, `LocalProtocolError`, `TooManyRedirects`) fail on the first attempt instead of burning the whole backoff. Everything else in the `impit.HTTPError` tree stays retryable, including a bare `HTTPError` from a body that ends mid-chunk and a `ProxyError` from a proxy answering CONNECT with a transient status.
- A failure while reading a streamed error response is classified and retried like a send failure, and the response is closed instead of being left open.
- `HttpResponse` comes from `typing_extensions`, whose runtime protocol check looks attributes up statically, so `isinstance(response, HttpResponse)` no longer buffers or rejects an unread streaming body.
- `try_import` only substitutes failed imports for the optional dependency it names, so an `ImportError` from the importing module or another dependency is no longer masked, and `__all__` omits what a missing extra never provided.

Split out of #1004, which adds the built-in HTTPX client on top of this contract.

*✍️ Drafted by Claude Code*

### DIFF
--- a/.rules.md
+++ b/.rules.md
@@ -60,7 +60,8 @@ Docstrings are written on sync clients and **automatically copied** to async cli
 
 ### HTTP Client Abstraction
 
-- `HttpClient`/`HttpClientAsync` — abstract base classes in `_http_clients/_base.py`
+- `HttpClient`/`HttpClientAsync` — base classes in `http_clients/_base.py` holding the shared request pipeline (retries,
+  timeouts, API errors); transports implement the `send_request`, error-classification, and lifecycle hooks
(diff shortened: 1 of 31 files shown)


vdusek:
This will be split into several smaller PRs

PR #1011, opened 2026-08-17
feat: Unify the request pipeline across HTTP clients

vdusek:
Moves `call`, the retry loop, and the per-attempt request handling from `ImpitHttpClient` and `ImpitHttpClientAsync` into `HttpClient` and `HttpClientAsync`, leaving Impit as a thin adapter over the transport hooks: `send_request`, `is_retryable_transport_error`, `is_timeout_error`, and `close()` / `aclose()` plus context managers. A custom client that overrides `call` keeps working unchanged and opts out of the shared pipeline. `StreamedLog` classifies timeouts through the transport-neutral `is_timeout_error` hook instead of importing Impit.

No behavior change for the built-in clients: retry classification stays exactly as on master (every `impit.HTTPError` is transient). The classification fixes follow in #1012.

Split out of #1006.

*✍️ Drafted by Claude Code*

### DIFF
--- a/.rules.md
+++ b/.rules.md
@@ -60,7 +60,8 @@ Docstrings are written on sync clients and **automatically copied** to async cli
 
 ### HTTP Client Abstraction
 
-- `HttpClient`/`HttpClientAsync` — abstract base classes in `_http_clients/_base.py`
+- `HttpClient`/`HttpClientAsync` — base classes in `http_clients/_base.py` holding the shared request pipeline (retries,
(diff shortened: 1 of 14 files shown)


vdusek:
Superseded by #1022 — the same pipeline unification rebased onto current master (which already carries the retry-classification fixes from #1019/#1020) and squashed into a single commit.

*✍️ Drafted by Claude Code*

PR #1012, opened 2026-08-17
fix: Fail fast on permanent transport errors and retry streamed error-read failures

vdusek:
- Transport errors a retry cannot fix (`UnsupportedProtocol`, `LocalProtocolError`, `TooManyRedirects`) fail on the first attempt instead of burning the whole backoff. Everything else in the `impit.HTTPError` tree stays retryable, including a bare `HTTPError` from a body that ends mid-chunk and a `ProxyError` from a proxy answering CONNECT with a transient status.
- A failure while reading a streamed error response is classified and retried like a send failure, and the response is closed instead of being left open.

Split out of #1006. Stacked on #1011.

*✍️ Drafted by Claude Code*

### DIFF
--- a/.rules.md
+++ b/.rules.md
@@ -60,7 +60,8 @@ Docstrings are written on sync clients and **automatically copied** to async cli
 
 ### HTTP Client Abstraction
 
-- `HttpClient`/`HttpClientAsync` — abstract base classes in `_http_clients/_base.py`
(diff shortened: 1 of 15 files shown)


vdusek:
Closing — this content already landed on master directly: #1019 (fail fast on permanent transport errors), #1020 (classify and retry streamed error-read failures), and #1021 (drop the unused retry helper). The pipeline unification continues in #1022, which preserves that behavior.

*✍️ Drafted by Claude Code*

PR #1020, opened 2026-08-18
fix: Classify and retry failures while reading a streamed error body

vdusek:
Before raising `ApifyApiError`, `_make_request` buffers the response so a streamed error body reaches the exception. That read sat outside the `try` block that classifies an attempt, so a failure there escaped the retry policy entirely: a non-retryable failure was retried through the whole backoff, and the response was left open.

The read now goes through the same classification as a failed send — retried when it is a transient transport failure, surfaced immediately when it is not — and the response is closed before the exception propagates.

Split out of #1006 and re-targeted at `master`, so the retry fixes can ship in a patch release ahead of the pipeline refactor in #1011.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/http_clients/_impit.py
+++ b/src/apify_client/http_clients/_impit.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import random
 import time
+from contextlib import suppress
 from datetime import timedelta
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Any, TypeVar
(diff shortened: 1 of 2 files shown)


src/apify_client/http_clients/_impit.py
```diff
@@ -232,9 +240,7 @@ def _make_request(
 
         except Exception as exc:
             logger.debug('Request threw exception', exc_info=exc)
-            if not _is_retryable_error(exc):
-                logger.debug('Exception is not retryable', exc_info=exc)
-                stop_retrying()
+            _stop_retrying_if_permanent(exc, stop_retrying=stop_retrying)
```
Pijukatel: Passing a fixture that controls that scoped variable... It works, but to me, it does not seem very readable.

Pijukatel (approved):

src/apify_client/http_clients/_impit.py
vdusek: ok, let's inline it

PR #1022, opened 2026-08-18
feat: Unify the request pipeline across HTTP clients

vdusek:
Moves `call`, the retry loop, and the per-attempt request handling from `ImpitHttpClient` and `ImpitHttpClientAsync` into `HttpClient` and `HttpClientAsync`, leaving Impit as a thin adapter over the transport hooks: `send_request`, `is_retryable_transport_error`, `is_timeout_error`, and `close()` / `aclose()` plus context managers. A custom client that overrides `call` keeps working unchanged and opts out of the shared pipeline. `StreamedLog` classifies timeouts through the transport-neutral `is_timeout_error` hook instead of importing Impit.

No behavior change for the built-in clients: the retry policy shipped in #1019 (permanent transport errors fail fast) and #1020 (a failed read of a streamed error body is classified like a failed send and the response is closed) is preserved — the classification now lives in the `is_retryable_transport_error` hook, and the error-body read handling in the shared pipeline.

Supersedes #1011, rebased onto current master and squashed. Originally split out of #1006.

Stack: **#1022 (this)** → #1004 (HTTPX client) → #1013 (docs).

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/http_clients/_base.py
+++ b/src/apify_client/http_clients/_base.py
@@ -1,13 +1,17 @@
 from __future__ import annotations
 
+import asyncio
 import json as jsonlib
 import logging
 import os
+import random
 import sys
-from abc import ABC, abstractmethod
+import time
+from contextlib import suppress
 from datetime import UTC, datetime, timedelta
+from http import HTTPStatus
 from importlib import metadata
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypeVar
 from urllib.parse import urlencode
 
 # `Protocol` comes from `typing_extensions`, not `typing`, because its runtime `isinstance` check looks attributes
@@ -25,17 +29,22 @@ from apify_client._consts import (
     MIN_COMPRESSION_SIZE,
 )
 from apify_client._docs import docs_group
-from apify_client._logging import LoggerOnce, logger_name
+from apify_client._logging import LoggerOnce, log_context, logger_name
 from apify_client._statistics import ClientStatistics
 from apify_client._utils.http import is_compressible_content_type
 from apify_client._utils.time import to_seconds
+from apify_client.errors import ApifyApiError
 from apify_client.http_compressors._gzip import GzipHttpCompressor
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncIterator, Iterator, Mapping
+    from collections.abc import AsyncIterator, Awaitable, Callable, Iterator, Mapping
+    from types import TracebackType
+    from typing import Self
 
     from apify_client.http_compressors._base import HttpCompressor
     from apify_client.types import JsonSerializable, Timeout
+
+T = TypeVar('T')
 
 logger = logging.getLogger(logger_name)
 logger_once = LoggerOnce(logger)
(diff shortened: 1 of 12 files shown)


src/apify_client/http_clients/_base.py
```diff
@@ -211,6 +236,24 @@ def _parse_params(params: dict[str, Any] | None) -> dict[str, Any] | None:
 
         return parsed_params
 
+    @staticmethod
+    def _is_body_worth_compressing(data: str | bytes | bytearray | None) -> bool:
+        """Whether this body clears the size threshold `_prepare_request_call` compresses at, cheaply.
```
Pijukatel: I had to read the first sentence several times, and I am still not sure I understand it correctly. Please rephrase it.

src/apify_client/http_clients/_base.py
```diff
@@ -211,6 +236,24 @@ def _parse_params(params: dict[str, Any] | None) -> dict[str, Any] | None:
...
+        Below the threshold nothing is ever compressed. At or above it the content type and a caller-supplied
+        `Content-Encoding` still decide, but checking those here would buy nothing - a body that turns out to be
+        already encoded only wastes the thread hop this answer guards.
+
+        The threshold is measured on encoded bytes, so a character count alone cannot decide a `str`. It is a
+        lower bound, so a `str` long enough in characters is long enough in bytes too. Below that the encoded
+        length decides, and the body is then under 4 KiB, so encoding it here is cheap.
+        """
+        if isinstance(data, str):
+            return len(data) >= MIN_COMPRESSION_SIZE or len(data.encode('utf-8')) >= MIN_COMPRESSION_SIZE
```
Pijukatel: `the body is then under 4 KiB, so encoding it here is cheap.`

I find it funny that body under 4KiB is too expensive for compression, but cheap enough for pointless encoding call :-)

Do we need such a strict boundary check here since it is only about optimization and not some functional requirements?

tests/unit/test_pluggable_http_client.py
```diff
@@ -142,7 +147,74 @@ async def call(
...
+        timeout: float | None,
+        stream: bool,
+    ) -> HttpResponse:
+        _ = stream
+        return await asyncio.to_thread(
+            _stdlib_fetch, method=method, url=url, headers=headers, content=content, timeout=timeout
+        )
+
+
+# -- Protocol conformance tests --
```
Pijukatel: I saw these comments were already in the file, but I am not a big fan of grouping the tests by free-floating comments. There is nothing enforcing the comments. I think that other methods of grouping tests are better, for example: standalone files, pytest markers, group by names, group into class...

Pijukatel (approved):
I see your PR attracted some of my comments to pre-existing code. Well, that is what you get for moving the code around :-D

tests/unit/test_pluggable_http_client.py
vdusek: Yeah I'm not a fan of them either. I even have a rule for it in the global `CLAUDE.md`. I dunno how this happened, maybe Claude saw a few other occurrences and added it here for consistency 🤷‍♂️.

src/apify_client/http_clients/_base.py
vdusek: yeah, so let's drop the dedicated str path

PR #1026, opened 2026-08-21
fix: Send request queue write fields under the names the API declares

vdusek:
`add_request` and `batch_add_requests` sent request fields snake_cased. The API declares its write bodies with `additionalProperties: false`, so it rejected the whole write with HTTP 400 — a request carrying `user_data`, `no_retry`, `retry_count`, `loaded_url`, `error_messages`, or `handled_at` could not be added at all. The identical dict passed to `update_request` worked:

```python
add_request     -> 400  {"uniqueKey": "k", "url": "...", "user_data": {...}, "no_retry": true}
update_request  -> 200  {"uniqueKey": "k", "url": "...", "userData":  {...}, "noRetry":  true}
```

The client was using `RequestDraft` (the spec's response-only schema for `unprocessedRequests`, which declares only `id`/`unique_key`/`url`/`method` and leaves the rest to `extra='allow'`) as its input model. The spec actually declares both add-request bodies as `RequestWithoutId`, now correctly generated as its own model since apify-docs#2774 fixed the spec bundler dropping its identity — no custom postprocessing needed. Upstream spec fix ([apify-docs#2774](https://github.com/apify/apify-docs/pull/2774)) is closed.

Two things to know:

- **Validation narrows.** The newly declared fields are validated instead of passed through, so `retry_count='abc'` or a naive `handled_at` now raise `ValidationError` instead of a 400 from the API.
- **`update_request`'s timestamp format changes** (`fa77d88`): `mode='json'` emits ISO 8601 where python mode emitted `2019-06-16 10:23:31.607000+00:00`. The API accepts both, so this is spec fidelity, not a fix — its own commit, revertible alone.

Integration tests round-trip every field through both add paths against the live API; they fail on master with `InvalidRequestError`.

*✍️ Drafted by Claude Code*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 ### 🐛 Bug Fixes
 
 - Require unique_key and url in RequestBase, add RequestWithoutId ([#1030](https://github.com/apify/apify-client-python/pull/1030)) ([28efd13](https://github.com/apify/apify-client-python/commit/28efd1388d18ab20ae95da24af415cdad3aae152)) by [@apify-service-account](https://github.com/apify-service-account)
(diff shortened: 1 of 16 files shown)


vdusek:
Btw. this is not an issue in the current SDK, because it uses the Crawlee's `Request` model with explicit `camelCase` aliases on fields.

Pijukatel (approved):

PR #1028, opened 2026-08-21
test: Make the mocked run status sequence independent of poll timing

vdusek:
`test_actor_call_redirect_logs_to_default_logger_sync` and its three siblings flake in CI, dropping `('Status: RUNNING, Message: Initial message', 20)` ([example](https://github.com/apify/apify-client-python/actions/runs/32504409582/job/96841227374)).

`StatusResponseGenerator` advanced the status every 5 requests, counted across every caller of the run endpoint — the two setup GETs, the watcher's 1-per-second polls, and `wait_for_finish`'s 4-per-second polling loop. `wait_for_finish` burned the "Initial message" window in roughly half a second, so the watcher's first poll had a ~0.5s margin to land inside it.

Plain requests and `waitForFinish` requests are now answered separately: plain requests walk the status sequence, and `waitForFinish` reports the run finished only once they have walked all of it. The first status covers three plain requests, which is `call`'s two setup GETs plus the watcher's own first poll in any order. With nothing watching there are no plain requests and the run reads as finished immediately, so `call(logger=None)` still returns at once.

Verified by injecting a delay before the watcher's first poll: on master the tests fail from 0.6s upward; with this change the two `call()`-based tests pass at 0.6s, 2s, 5s and 10s. (`test_redirect_status_message_sync` still fails past ~5s, but that is bounded by its own 1s `sleep`, not by the mock.)

*✍️ Drafted by Claude Code*

### DIFF
--- a/tests/unit/test_logging.py
+++ b/tests/unit/test_logging.py
@@ -84,12 +84,23 @@ _EXPECTED_MESSAGES_AND_LEVELS_WITH_STATUS_MESSAGES = (
 
 
 class StatusResponseGenerator:
-    """Generator for actor run status responses to simulate changing status over time."""
+    """Generator for actor run status responses to simulate changing status over time.
+
+    A status watcher polls the plain endpoint once a second while `wait_for_finish` polls it several times a second
+    with `waitForFinish`. Driving one sequence from both would make the messages a watcher observes depend on which
+    poll got there first, so the two are answered separately: plain requests walk the sequence, and `waitForFinish`
+    reports the run finished only once they have walked all of it.
+    """
+
+    _requests_for_first_status = 3
+    """Plain requests answered with the first status before the sequence moves on.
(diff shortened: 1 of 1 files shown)


vdusek:
Closing - the flake this targeted was in the client, not the mock. In the CI run cited above, `Status: RUNNING, Message: Initial message` appears in captured stderr but not in `caplog`, so the watcher's poll did land inside the first-status window; the record was dropped by the logger rebuild that #1036 fixed.

The mock's timing coupling is real but latent, and the trade here isn't worth it: holding the run open until the plain sequence is walked means a stalled watcher hangs `wait_for_finish` indefinitely instead of failing fast, and `_requests_for_first_status = 3` hardcodes how many setup GETs `call()` happens to make.

*✍️ Drafted by Claude Code*

PR #1046, opened 2026-09-02
fix: Name the extra to install in optional-dependency import errors

vdusek:
Accessing a symbol guarded by `try_import` without its extra installed raises an `ImportError` that now also names the extra to install:

```
No module named 'httpx2'. Install the optional 'httpx' extra to use it: pip install 'apify-client[httpx]'
```

`try_import` takes `extra_name` next to `dependency_name`, and both guarded sites (`httpx`, `brotli`) pass it.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_utils/try_import.py
+++ b/src/apify_client/_utils/try_import.py
@@ -19,11 +19,12 @@ class ImportState:
 
 
 @contextmanager
-def try_import(module_name: str, *symbol_names: str, dependency_name: str) -> Iterator[ImportState]:
+def try_import(module_name: str, *symbol_names: str, dependency_name: str, extra_name: str) -> Iterator[ImportState]:
     """Context manager to attempt importing symbols into a module.
 
-    If the named optional dependency is missing, the symbols are replaced with `FailedImport` objects. Import errors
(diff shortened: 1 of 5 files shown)


vdusek:
Open question, do we want the extra to be named `httpx` or `httpx2`?

```
No module named 'httpx2'. Install the optional 'httpx' extra to use it: pip install 'apify-client[httpx]'
```

or

```
No module named 'httpx2'. Install the optional 'httpx2' extra to use it: pip install 'apify-client[httpx2]'
```

I'm slightly more into naming the extra according to the package, so `httpx2` (in this case, we will align this in Crawlee in v2).

cc @Pijukatel , @Mantisus

Mantisus:
> Open question, do we want the extra to be named `httpx` or `httpx2`?

I prefer `httpx2`. When `httpx` releases version 1.0 with a new API, keeping `httpx` in the extra will confuse users.

vdusek:
> I prefer httpx2. When httpx releases version 1.0 with a new API, keeping httpx in the extra will confuse users.

Good point.

Pijukatel (approved):

vdusek:
Let's do the `httpx2` rename in another PR

PR #1059, opened 2026-09-10
fix: Stop dataset iterators from skipping items when unwind is used

### LINKED ISSUE #1058: Dataset iterators skip items when `unwind` returns more items than the rows scanned
## Summary

`iterate_items` advances the dataset offset by `max(scanned_rows, len(items))`. With `unwind`, a page returns more items than the rows it scanned, so the offset jumps past rows the next request never reads and those items are silently dropped.

## Where

The `max()` appears in two layers, four sites:

| File | Line | Code |
| --- | --- | --- |
| `_resource_clients/dataset.py` | 209 (sync), 768 (async) | `count=max(int(response.headers['x-apify-pagination-count']), len(items))` |
| `_pagination.py` | 68 (sync), 99 (async) | `page_scanned = max(getattr(current_page, 'count', 0), len(current_page.items))` |

The API applies `offset` and `limit` to the dataset's rows first and shapes the result afterwards. `x-apify-pagination-count` is set from `min(total - offset, limit)` before the transform stream runs, so it reports the rows scanned and never reflects `unwind` or the filters. Filters (`clean`, `skip_empty`, `skip_hidden`) leave `len(items)` below that number, `unwind` leaves it above. The `max()` handles the first case and breaks the second.

## Reproduction

Driving the real iterator with a 4-row dataset where every row unwinds into three items, at `chunk_size=2`:

```python
from dataclasses import dataclass

from apify_client._pagination import get_items_iterator

TOTAL_ROWS = 4
PARTS = ['a', 'b', 'c']


@dataclass
class Page:
    items: list[dict]
    count: int


calls = []


def callback(*, limit: int | None = None, offset: int | None = None) -> Page:
    offset = offset or 0
    calls.append(offset)
    rows = [{'id': i} for i in range(offset, min(offset + (limit or TOTAL_ROWS), TOTAL_ROWS))]
    # The API unwinds each row into three items after offset/limit picked the row window.
    items = [{**row, 'part': p} for row in rows for p in PARTS]
    # dataset.py builds the page as max(header, len(items)); the header is len(rows).
    return Page(items=items, count=max(len(rows), len(items)))


got = list(get_items_iterator(callback, chunk_size=2))
print('requested offsets:', calls)
print('yielded items:', len(got), 'of', TOTAL_ROWS * len(PARTS))
```

```
requested offsets: [0, 6]      # should be [0, 2]
yielded items: 6 of 12
```

Rows 2 and 3 are never requested. At the default `chunk_size=1000`, a 5000-row dataset with a 3x unwind yields about 2000 rows.

Nothing bounds the over-advance: `get_items_iterator` documents that "the `total` field is intentionally not consulted, because it can change between calls", so the offset runs off the end of the dataset and the loop stops on the resulting empty page.

## Affected versions

Both `max()` sites arrived with #771 and have shipped since v3.0.0. #964 changed only the termination condition and left the bookkeeping alone.

## Suggested fix

Follow the scanned count alone, which is right in both directions, and fall back to `len(items)` only for responses that expose no scanned count (the collection endpoints). apify-client-js took this route in [#1044](https://github.com/apify/apify-client-js/pull/1044) after the same finding in review.

That forces a decision on the public `DatasetItemsPage.count`, which is currently `max(scanned, returned)` while its docstring says "Number of items in this page". apify-client-js keeps `count == len(items)` and carries the scanned number out of band, so the two clients currently disagree on this field.

## Two related items

- There is no `unwind` coverage anywhere in `tests/`, which is why this went unnoticed.
- The `get_cursor_iterator` docstring says server-side filters "can drop every item on a page while a live cursor still points at more data". The API rules that out. For the request queue, `getNextCursor()` returns a cursor only when `pushedItemsCount >= limit` and `limit` is validated above zero. For the key-value store, `nextExclusiveStartKey` is `last(keys)?.key`, which is absent on an empty page. `test_cursor_iterator_continues_past_fully_filtered_page` covers a state the API cannot produce. The behavior is harmless, the justification is wrong.

*✍️ Drafted by Claude Code*

vdusek:
## What was wrong

`iterate_items` advanced the dataset offset by `max(scanned_rows, len(items))`. With `unwind`, a page returns more
items than the rows it scanned, so the offset jumped past rows the next request never read. At the default
`chunk_size`, a 5000-row dataset with a 3x unwind yielded about a third of it.

## The fix

The offset now advances by at most the number of rows the call asked for. The items endpoint applies `offset` and
`limit` to the rows first and shapes the result afterwards, and it passes no `maxLimit` (unlike the collection list
routes), so a page never covers more rows than the `limit` it was sent. Capping the advance there keeps an unwound
page from running past the window it actually read.

The `max()` stays underneath the cap on purpose. `x-apify-pagination-count` comes from the dataset's `itemCount`,
which the API increments through a throttled write while the items themselves stream fresh. Following the header
alone, which is what [apify-client-js#1044](https://github.com/apify/apify-client-js/pull/1044) does and what #1058
suggested, would truncate `iterate_items` right after `push_items` for anything over one page. The JS iterator
bounds itself by `total` and so is exposed to that lag either way; this one consults neither, which is why the two
clients end up with different fixes.

`DatasetItemsPage.count` keeps its current value of `max(header, len(items))`. That is not the no-op it looks like:
it is what makes `count` usable right after a push, and two integration assertions depend on it. Its docstring
changes, along with the `limit` docs on `iterate_items` and `chunk_size` on both twins, since all of them promised
items where the field and the parameters have always counted scanned rows.

One case stays imperfect. On a final short page with `unwind`, the advance is still the full `limit` rather than the
rows the page covered, so rows appended by a concurrent push after that page can be missed. Reaching them needs the
raw header, which `list_items` folds into `count`. It is strictly better than before, and the docstring says so
rather than claiming the overshoot is free.

## Behavior change worth a changelog line

`iterate_items(limit=N)` with `unwind` and a `chunk_size` below `N` now walks all N rows instead of stopping after
the first page, so it can yield several times more items for the same `limit`. That is the fix working, but it is
user-visible.

## Tests

- `unwind` reached the pagination tests for the first time. The fake API now splits each row into three items after
  the row window is picked, exactly as the transform stream does, and three cases iterate through it.
- The fake API used to cap every page at 1000 items "mirroring the real API", which is only true of the collection
  endpoints. It now applies the requested limit verbatim on the items endpoint, which is what let a `chunk_size`
  above that cap be covered at all.
- A page whose `count` lags behind its items has a test of its own. Nothing pinned that before, and it is the
  invariant the `max()` rests on.
- `test_dataset_iterate_items_unwound` covers the whole thing against the live API. It has not run locally, so CI
  is its first real execution.

## One thing found while verifying

`get_cursor_iterator` justified its termination rule with filters that "can drop every item on a page while a live
cursor still points at more data". The API rules that out: the key-value store returns the last key of the page as
the next cursor, and the request queue returns a cursor only once a page came back full, so neither can hand back a
cursor that outlives its page. The behavior is fine and unchanged; the docstring now says why it holds.

Closes #1058

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_client/_pagination.py
+++ b/src/apify_client/_pagination.py
@@ -48,8 +48,8 @@ def get_items_iterator(
 
     Args:
         callback: Function returning a single page of items.
-        limit: Maximum total number of items scanned across all pages, counted in rows on the dataset items endpoint
-            where `unwind` can turn one row into several yielded items. `None` or `0` means no limit.
+        limit: Maximum total number of rows scanned across all pages. On the dataset items endpoint `unwind` can
+            turn one row into several items, so more items than this can be yielded. `None` or `0` means no limit.
         offset: Starting offset for the first page.
         chunk_size: Per-page cap, sent to the API as its `limit`. `None` or `0` lets the API decide.
     """
@@ -132,9 +132,9 @@ def get_cursor_iterator(
 
     Cursor pagination is restricted to the two API responses that expose it: `ListOfKeys` (for key-value store keys) and
     `ListOfRequests` (for request queue requests). Iteration ends when the next cursor is `None` or the user-requested
-    `limit` is reached; an empty page on its own does not stop it. Both endpoints derive the cursor from the page they
-    return - the key-value store hands back the last key of the page, the request queue hands back a cursor only once a
-    page came back full - so termination can rest on the cursor alone.
+    `limit` is reached; an empty page on its own does not stop it. Neither endpoint can hand back a cursor that
+    outlives its page - the key-value store returns the last key of the page, and the request queue returns a cursor
+    only once a page came back full - so termination can rest on the cursor alone.
 
     Args:
         callback: Function returning a single page of items. Receives `cursor` and `limit` kwargs.
--- a/src/apify_client/_pagination.py
+++ b/src/apify_client/_pagination.py
@@ -19,9 +19,10 @@ The value of 1000 keeps backwards compatibility with the previous fixed cache si
 class HasItems(Protocol[T]):
     """Structural contract for a single page of results from a paginated API endpoint.
 
-    Implementations must expose `items`. They may optionally expose `count` - the number of items scanned by the API for
-    this page, which can exceed `len(items)` when filters drop items from the response. The iterator helpers consult
-    `count` opportunistically via `getattr` for offset bookkeeping and fall back to `len(items)` when it is absent.
+    Implementations must expose `items`. They may optionally expose `count` - the number of rows the API scanned to
+    produce this page, which `len(items)` can land below (filters drop items) or above (`unwind` splits one row into
+    several items). The iterator helpers consult `count` opportunistically via `getattr` for offset bookkeeping and
+    fall back to `len(items)` when it is absent.
     """
 
     items: list[T]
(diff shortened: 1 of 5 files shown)


src/apify_client/_pagination.py
```diff
@@ -130,10 +132,9 @@ def get_cursor_iterator(
 
     Cursor pagination is restricted to the two API responses that expose it: `ListOfKeys` (for key-value store keys) and
     `ListOfRequests` (for request queue requests). Iteration ends when the next cursor is `None` or the user-requested
-    `limit` is reached. Emptiness alone does not stop iteration: server-side filters (such as the request-queue state
-    `filter`) can drop every item on a page while a live cursor still points at more data, so termination relies on the
-    cursor, not on whether a page returned items. Unlike offset responses, cursor responses expose no scanned-item
-    `count`, so `count` cannot be used to detect a fully-filtered page here.
+    `limit` is reached; an empty page on its own does not stop it. Neither endpoint can hand back a cursor that
```
Pijukatel: I find this doc update less readable, and I am not sure if it's saying the same thing as before or if it tries to tell something different.

What is a `cursor that outlives its page`?

vdusek: OK, I completely reworked this docstring. Hope it's better now.

Pijukatel (approved):

PR #1060, opened 2026-09-11
feat: Stream request bodies from files, iterators, and responses

### LINKED ISSUE #972: Support streaming request bodies to avoid buffering large uploads in memory

vdusek:
## What changed

`set_record`, the `run_input` of `start` / `call` / `metamorph`, and `HttpClient.call(data=...)` now accept a file-like object, an iterator of `bytes` or `str` chunks, or a streamed `HttpResponse`. The body goes out chunked and uncompressed, so one Actor's `OUTPUT` record can become another Actor's input without the orchestrating process holding it in memory. The async client also takes async iterators and `aiofiles`-style objects; the sync client rejects those with a `TypeError` naming `ApifyClientAsync`.

A file-like object is read through `read(chunk_size)`. A `read` that takes no size is called once and its result sent as a single chunk, so a duck-typed reader of that shape keeps working and keeps its old memory profile.

## Retries

Only a seekable file-like source is retried, sought back to its starting position before each attempt. Anything else gets a single attempt. An error raised inside the body iterator surfaces as itself with the transport error as its cause, since Impit reports it as a bare `impit.HTTPError` that the classifier would treat as transient.

## Compression

A file-like `set_record` value is no longer compressed, so uploading from a file handle sends more bytes over the wire, which Apify bills as data transfer. Passing `file.read()` keeps the old behavior. Not breaking: the call keeps working, and the extra data transfer belongs in the changelog rather than in a major bump.

A streamed value that declares a `content_encoding` is taken at its word, since the client only sees its bytes as they go out. A `str`, a JSON-serializable object, and a text-mode file are still rejected.

## Custom transports

`send_request` now receives `bytes | Iterator[bytes] | None` (async: `AsyncIterator[bytes]`), and `call` accepts the new source types. A transport typed against `bytes | None` only sees an iterator once one of its own callers opts into streaming, so this isn't breaking either.

## Docs

A "Streaming uploads" section on the streaming concept page, plus updates to the compression and HTTP client pages, the custom transport examples, and the README. The guide on chaining Actors into a pipeline is a separate PR.

## Tests

`StreamedRequestBody` on its own, the retry pipeline with fake transports, and `set_record` end to end over Impit and HTTPX2. The two integration tests (3 MiB upload, record piped between stores) weren't run locally.

## Issues

- Closes #972

*✍️ Drafted by Claude Code*


src/apify_client/http_clients/_base.py
```diff
@@ -320,23 +325,36 @@ def _prepare_request_call(
             if self._get_header(headers, 'content-type') is None:
                 headers['Content-Type'] = 'application/json'
 
+        # A body the caller built itself carries its own chunk size and rewind position, and wrapping it again
+        # would read it as a response and drop both.
+        if isinstance(data, StreamedRequestBody):
+            return (headers, self._parse_params(params), data)
+
+        if StreamedRequestBody.is_source(data):
```
Pijukatel, line 333: `StreamedRequestBody.is_source(data)` reads more like `is data an existing source of StreamedRequestBody`

Maybe it should have a name that implies that the method is testing a possibility, not the existing reality.

I am bad with naming, but just to give a hint in which direction I think it should go
`StreamedRequestBody.is_possible_source(data)`
`StreamedRequestBody.is_compatible_source(data)`
`StreamedRequestBody.can_use_as_source(data)`
...

src/apify_client/http_clients/_streamed_body.py
```diff
@@ -0,0 +1,257 @@
...
+
+        These are a streamed `HttpResponse` (anything with a callable `iter_bytes`), a file-like object (anything
+        with a callable `read`), and an iterator or async iterator of byte chunks. A `str`, `bytes`, `bytearray`, or
+        a container such as a `list` or `dict` is not a source, even though some of them can be iterated.
+
+        A `StreamedRequestBody` matches on its own `iter_bytes`, which is how a hand-built body reaches the request
+        pipeline untouched. The constructor refuses one, so a caller that builds a body from what this accepts has
+        to check for an existing body first.
+        """
+        return (
```
Pijukatel, line 114: This seems too restrictive for all iterables. (We want to keep some special iterables - like strings, lists, ... out of streaming, but is there really a reason to keep all of them out?l)

For example (simplified example just to demonstrate the point)

```
class Chunked:
    """Wraps a string and yields it in pieces. Only __iter__, nothing else."""

    def __init__(self, text: str, size: int = 10) -> None:
        self.text, self.size = text, size

    def __iter__(self):
        for i in range(0, len(self.text), self.size):
            yield self.text[i : i + self.size]


client.actor('E2jjCZBezvAZnX8Rb').start(
    run_input=Chunked('{"message": "Hello world!"}'),
    content_type='application/json',
)
```

On your PR it is serialized as 
"<__main__.Chunked object at 0x7f97fb28a8d0>"

I asked to create a commit with an example implementation that would also support these:
https://github.com/apify/apify-client-python/commit/7b34e44aa4ec0cf054c4f36c039ab54d1dc40ecc

src/apify_client/http_clients/_streamed_body.py
```diff
@@ -0,0 +1,257 @@
...
+        self._read_takes_size = True
+        self._sync_chunks: Callable[[], Iterable[Any]] | None = None
+        self._async_chunks: Callable[[], AsyncIterator[Any]] | None = None
+
+        # Set for a seekable file-like source, the only kind that can be sent more than once.
+        self._seek: Callable[[int], Any] | None = None
+        self._start: int | None = None
+
+        # A response is recognized first, so its `read` is never touched - on an unread streaming response that
+        # either raises or buffers the whole body.
```
Pijukatel, line 75: It does not feel very robust.

The detection in StreamedRequestBody is a chain of getattr probes. That works for the standard library, but this code takes arbitrary user objects, and for those the outcome depends on which method names they happen to have. 

`_accepts_chunk_size` is going even further with guessing.

So for user-defined code or libraries we have not considered, this can behave in completely unpredictable ways.

